PR #12, opened 2015-07-05
Readme updates

doomspork:
Hey @hassox, thanks for Guardian!  After going through the docs and found two small changes:

+ Fix a "event" => "even" typo
+ Move the `EnsureSession` example from `EnsurePermissions` to `EnsureSession`

Looking forward to your feedback


hassox:
Thanks @doomspork!

PR #18, opened 2015-08-20
Use sub claim in README to load resource

bbhoss:
In the above example, the serializer is designed to only take input from the sub field. While you could also implement a pattern to load a resource from the full claim set, I think this is the intended and most common way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ end
 Accessing the resource from a set of claims:
 
 ```elixir
-case Guardian.serializer.from_token(claims) do
+case Guardian.serializer.from_token(claims.sub) do
   { :ok, resource } -> do_things_with_resource(resource)
   { :error, reason } -> do_things_without_a_resource(reason)
 end


hassox:
This would be a breaking change. See the Guardian.Plug.LoadResource plug. 

Can you give an example of when the full claims would be useful? I'm pretty open if there's a good use case. We'd need to work out a way to move forward with such a change.

bbhoss:
I'm not sure what you think I'm proposing we change. I was just correcting what I thought was a documentation issue. Your example serializer depends on only the sub being passed in instead of the full claim.

hassox:
The current fom_token method doesn't expect the claims, only the entry in the sub field. If you look at the load resource plug:

    case Guardian.serializer.from_token(Dict.get(claims, "sub")) do

The sub field is the only one that is passed along. The example in the readme is correct for how the library works at the moment.

hassox:
Doh! I totally read this in reverse. Thanks for the patience explaining it!

PR #40, opened 2015-09-06
Replace generic Dict calls in favour of Map

andreaazzini:
Closes https://github.com/hassox/guardian/issues/26.

### DIFF
--- a/lib/guardian/channel.ex
+++ b/lib/guardian/channel.ex
@@ -36,7 +36,7 @@ defmodule Guardian.Channel do
   """
   defmacro __using__(opts) do
     opts = Enum.into(opts, %{})
-    key = Dict.get(opts, :key, :default)
+    key = Map.get(opts, :key, :default)
 
     quote do
       def join(room, auth = %{ "guardian_token" => jwt }, socket) do
@@ -48,7 +48,7 @@ defmodule Guardian.Channel do
       defp handle_guardian_join(room, jwt, params, socket) do
         case Guardian.decode_and_verify(jwt, params) do
           { :ok, claims } ->
-            case Guardian.serializer.from_token(Dict.get(claims, "sub")) do
+            case Guardian.serializer.from_token(Map.get(claims, "sub")) do
               { :ok, resource } ->
                 authed_socket = socket
                 |> assign(Guardian.Keys.claims_key(unquote(key)), claims)
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -68,8 +68,8 @@ defmodule Guardian.Plug do
   """
   @spec sign_in(Plug.Conn.t, any, atom | String.t, Map) :: Plug.Conn.t
   def sign_in(conn, object, type, claims) do
-    the_key = Dict.get(claims, :key, :default)
-    claims = Dict.delete(claims, :key)
+    the_key = Map.get(claims, :key, :default)
+    claims = Map.delete(claims, :key)
 
     case Guardian.encode_and_sign(object, type, claims) do
       { :ok, jwt, full_claims } ->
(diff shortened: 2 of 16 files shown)


lib/guardian/channel.ex
```diff
@@ -36,7 +36,7 @@ defmodule Guardian.Channel do
   """
   defmacro __using__(opts) do
     opts = Enum.into(opts, %{})
-    key = Dict.get(opts, :key, :default)
+    key = Map.get(opts, :key, :default)
```
hassox: This may not be a map so should stay dict.

lib/guardian/plug/ensure_permissions.ex
```diff
@@ -13,16 +13,16 @@ defmodule Guardian.Plug.EnsurePermissions do
   """
   def init(opts) do
     opts = Enum.into(opts, %{})
-    on_failure = Dict.get(opts, :on_failure)
-    key = Dict.get(opts, :key, :default)
-    perms = Dict.delete(opts, :on_failure) |> Dict.delete(:key)
+    on_failure = Map.get(opts, :on_failure)
```
hassox: May not be a map. This should stay a Dict.

lib/guardian/plug/ensure_authenticated.ex
```diff
@@ -15,7 +15,7 @@ defmodule Guardian.Plug.EnsureAuthenticated do
   @doc false
   def init(opts) do
     opts = Enum.into(opts, %{})
-    case Dict.get(opts, :on_failure) do
+    case Map.get(opts, :on_failure) do
```
hassox: May not be a map. Should stay a Dict.

lib/guardian/plug.ex
```diff
@@ -50,8 +50,8 @@ defmodule Guardian.Plug do
   """
   @spec sign_in(Plug.Conn.t, any, atom | String.t, Map) :: Plug.Conn.t
   def sign_in(conn, object, type, claims) do
-    the_key = Dict.get(claims, :key, :default)
-    claims = Dict.delete(claims, :key)
+    the_key = Map.get(claims, :key, :default)
```
hassox: This may not be a map. This should stay a Dict.

lib/guardian/utils.ex
```diff
@@ -3,7 +3,7 @@ defmodule Guardian.Utils do
   @doc false
   def stringify_keys(nil), do: %{}
   def stringify_keys(map) do
-    Enum.reduce(Dict.keys(map), %{}, fn(k,acc) -> Dict.put(acc, to_string(k), map[k]) end)
+    Enum.reduce(Map.keys(map), %{}, fn(k,acc) -> Map.put(acc, to_string(k), map[k]) end)
```
hassox: This first one may not be a map. The one inside the anon function is for sure, but the other... not so sure.

hassox:
LGTM. Just a couple of places where there could be either a map or KW list.

andreaazzini:
I think the problem with those two places reside in the nature of `stringify_keys` itself. If we handled its input data so that it is not `nil` in the first place, we'd be completely fine using `Map`. What do you think about this?

hassox:
The stringify keys could be getting a keyword list from arguments. If we make sure that those places are all good I think we should be ok. We'll just have to Enum.into a map before calling stringify keys.

lib/guardian/utils.ex
lexmag: I think for this case `Enum.reduce(map, %{}, fn({key, val},acc) -> Map.put(acc, to_string(key), val) end)` would be way better.

andreaazzini: This looks good to me. @hassox any thought on this?

hassox:
This is aging. I've been cleaning the Dict calls up as I come across them. Please update this PR if you want me to bring it in. :+1:

hassox:
sorted in https://github.com/ueberauth/guardian/pull/77

PR #54, opened 2015-10-06
fetch session before putting stuff in it

nolman:
I believe one needs to fetch_session before one put_session's

https://github.com/phoenixframework/phoenix/issues/173


hassox:
I've added a new function to Guardian.Plug to help with signing into api endpoints. 

This new function does the same as the Guardian.Plug.sign_in function with the exception that it doesn't try and put things in the session.

    Guardian.Plug.api_sign_in(<same as sign_in)

This sets the JWT, claims and resource into the connections assigns. The decision of if you should use the session variant or not should fall within the pipeline that you're expecting to use for the request.

nolman:
Thanks dude, you rock

PR #58, opened 2015-10-31
Fix two bugs with api_sign_in

### LINKED ISSUE #55: Unable to get value from conn.assigns using Guardian.Plug.claims
I don't think checking the value in conn.assigns dict returns a tuple. all the documentation i can find points now to returning just the value or nil. Since `Guardian.Plug.claims` is checking for this tuple, https://github.com/hassox/guardian/blob/master/lib/guardian/plug.ex#L161, it's always failing for me with the `:no_session` error case. This breaks me checking claims and the various plugs like `EnsureAuthenticated` that use it. I'm using Plug v1.0.2 and I noticed the test suite is set up for v1.0.0 but i can't see the difference in return values for it between the versions. Is there something I'm missing?

doomspork:
Hi @hassox.  I started digging into the `api_sign_in` code after seeing #55 and along the way I found two bugs which the PR fixes including the original issue:

+ The parameters for `set_claims` during errors were mixed up.  I re-ordered them and added a test case.
+ Neither `sign_in` or `api_sign_in` were passing the expected tuple to `set_claims`.  Closes #55

### DIFF
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -136,7 +136,7 @@ defmodule Guardian.Plug do
         |> set_current_token(jwt, the_key)
         |> Guardian.hooks_module.after_sign_in(the_key)
 
-      { :error, reason } -> set_claims(conn, base_key(the_key), { :error, reason }) # TODO: handle this failure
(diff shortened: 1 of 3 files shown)


hassox:
Whoa thankyou!

doomspork:
Thank __you__ for Guardian!

PR #67, opened 2015-11-27
Move on_failure to a handler module

hassox:
The on\_failure was pretty verbose. This PR moves that to a simple handler that should define 
`:unauthenticated` and `:unauthorized` function instead.

ping @chrismccord

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ If one is not found, this does nothing.
 ### Guardian.Plug.EnsureAuthenticated
 
 Looks for a previously verified token. If one is found, continues, otherwise it
-will call the `:on_failure` function.
+will call the `:unauthenticated` function of your handler.
 
 When you ensure a session, you must declare an error handler. This can be done
 as part of a pipeline or inside a pheonix controller.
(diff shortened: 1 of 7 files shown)


doomspork:
This is a great change @hassox :+1:

README.md
```diff
@@ -101,25 +101,41 @@ as part of a pipeline or inside a pheonix controller.
...
+  plug Guardian.Plug.EnsurePermissions, handler: MyApp.MyAuthErrorHandler, default: [:read, :write]
 end
 ```
 
+The error handler may be defined inline or in the main configuration.
+
+```elixir
+config :guardian, Guardian,
+  handler: MyApp.MyAuthErrorHandler
+  # …
```
chrismccord: This is global, so I am :-1:  on it. The user can define their own handler module, then plug that anywhere they want.

```elixir
defmodule MyApp.EnsureAuthenticated do
  use Plug.Builder
  plug Guardian.Plug.EnsureAuthenticated, handler: __MODULE__

  def unauthenticated(conn) do
    ...
  end
end


defmodule MyApp.MyController do
  plug MyApp.EnsureAuthenticated
end
```

README.md
```diff
@@ -152,7 +168,7 @@ From here, you can either EnsureAuthenticated in your pipeline, or on a per-cont
 defmodule MyApp.MyController do
   use MyApp.Web, :controller
 
-  plug Guardian.Plug.EnsureAuthenticated, on_failure: { MyApp.MyHandler, :unauthenticated }
+  plug Guardian.Plug.EnsureAuthenticated, handler: MyApp.MyAuthHandler
 end
```
chrismccord: So to be clear, explicit handler like this is fine, it's the global config I am against. Otherwise looks good!

Issue #84, opened 2016-01-09
Recompile on permissions update?


skosch:
Hi,

forgive the somewhat unrelated question, but as a newcomer to Elixir, I just pulled my hair out for way too long because my permissions weren't working ... until I realized that `mix` actually compiles the Map in `config.exs` directly into the `@perms` attribute, and that said compilation isn't automatically re-triggered when my configuration changes. In other words, I have to delete `_build/guardian` after updating my permissions or else my app won't work.

If this should happen automatically, then something must be missing in my setup, and I'd be grateful for a pointer to a working example.

hassox:
oh that's a good point. The GuardianPermissions module is dynamically generated. I'm not sure how to watch for changes in the config and re-generate. That's an interesting challenge though.

skosch:
Hm. I guess it's not that big of a deal in practice, but maybe a note in the docs would be helpful to save the next guy a few hours :)

hassox:
Quick update on this. I've looked into getting some code reloading happening, and while it's very possible, it still doesn't help with partial compilations like tests. Only in dev. I'm going to take a stab at refactoring the permissions module to be a more run-time set of functions rather than compile time definitions.

skosch:
Sweet, thank you! Looking forward to what you'll come up with. I'll keep the issue open until then.

sschneider1207:
It would be nice to have a notice about this in the section of the readme on permissions, I just had this same problem and couldn't figure out what was wrong for the life of me.

PR #88, opened 2016-01-17
Do not compile permissions. Let them be fetched at runtime

hassox:
Fixes https://github.com/ueberauth/guardian/issues/84#issuecomment-171517247

### DIFF
--- a/lib/guardian/permissions.ex
+++ b/lib/guardian/permissions.ex
@@ -86,53 +86,7 @@ defmodule Guardian.Permissions do
   """
   use Bitwise
 
-  perms = Enum.into(Guardian.config(:permissions, %{}), %{})
-  @perms perms
   @max -1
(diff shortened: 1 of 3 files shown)


lib/guardian/permissions.ex
```diff
@@ -86,68 +86,26 @@ defmodule Guardian.Permissions do
...
 
   @doc """
   Fetches the list of known permissions for the given type
   """
   @spec available(atom) :: List
-  def available(type), do: Map.get(@perms, type, [])
+  def available, do: available(:default)
+  def available(type) when is_binary(type) do
+    try do
+      available(String.to_existing_atom(type))
```
doomspork: I think you should use pipes here:

```elixir
type
|> String.to_existing_atom
|> available
```

PR #89, opened 2016-01-19
Refactored Guardian.encode_and_sign/3

aaronrenner:
I was having a difficult time trying to figure out how the claims are built, so I did some refactoring on the `Guardian.encode_and_sign/3` function. I isolated the claim building functionality into a single private method and wrapped calling the callback hooks in some easier to call method names. All of the tests pass, and I added an additional test for when the serializer fails to generate a token for a given model.

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -52,29 +52,17 @@ defmodule Guardian do
   """
   @spec encode_and_sign(any, atom | String.t, Map) :: { :ok, String.t, Map } | { :error, atom } | { :error, String.t }
   def encode_and_sign(object, type, claims) do
-    claims = stringify_keys(claims)
-    perms = Map.get(claims, "perms", %{})
-    claims = Guardian.Claims.permissions(claims, perms) |> Map.delete("perms")
-
-    case Guardian.serializer.for_token(object) do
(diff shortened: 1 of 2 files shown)


aaronrenner:
On a side note, I did notice that the 3 levels of nested case statements is pretty rough to read. What would you think if I made another pull request that reworks encode_and_sign to use a with statement like this?

```elixir
def encode_and_sign(object, type, claims) do
  with {:ok, claims} <- build_claims(object, type, claims),
       {:ok, {object, type, claims}} <- call_before_encode_and_sign_hook(object, type, claims),
       {:ok, jwt} <- encode_claims(claims),
       call_after_encode_and_sign_hook(object, type, claims, jwt),
       do: { :ok, jwt, claims }
end
``` 

The only drawback is it would require that the Elixir dependency be bumped to 1.2.

hassox:
This is nice! thankyou. I'm not sure about the with call. I've been wondering that myself but the bump to 1.2. I think it's probably ok to require an upgrade to 1.2. I don't imagine there's much left that is on 1.1 at this point. I'd be down for the bump to 1.2 on a different PR which we can then bump as a definitive version number.

PR #93, opened 2016-01-25
Bump Elixir dependency to 1.2

aaronrenner:
This pull request bumps the elixir dependency to 1.2 so we can refactor nested case statements to use the new `with` syntax.

I do have a couple of questions:

1. `decode_and_verify/2` originally called `stringify_keys` twice in the following code:

  ```elixir
  def decode_and_verify(jwt, params) do
    params = stringify_keys(params)
    if verify_issuer?, do: params = Map.put_new(params, "iss", issuer)
    params = stringify_keys(params)

    # Rest of method ommitted
  end
  ```
  The second call to `stringify_keys` seemed to be redundant since the key that's being inserted is already a string. I removed the second `stringify_keys` call when I extracted the `build_expected_claims` function, but I was wondering if there is a reason that I shouldn't have removed it.
2. I also saw that there was a `try`/`rescue` block in `decode_and_verify/2`. What is the purpose of this and can it be removed?

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -52,20 +52,11 @@ defmodule Guardian do
   """
   @spec encode_and_sign(any, atom | String.t, Map) :: { :ok, String.t, Map } | { :error, atom } | { :error, String.t }
   def encode_and_sign(object, type, claims) do
-    case build_claims(object, type, claims) do
-      { :ok, claims_for_token } ->
-        case call_before_encode_and_sign_hook(object, type, claims_for_token) do
-          { :ok, { resource, type, claims_from_hook } } ->
-            case encode_claims(claims_from_hook) do
-              { :ok, jwt } ->
-                call_after_encode_and_sign_hook(resource, type, claims_from_hook, jwt)
-                { :ok, jwt, claims_from_hook }
-              { :error, reason } -> { :error, reason }
-            end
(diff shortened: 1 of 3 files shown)


doomspork:
I'm not sure everyone is on 1.2 (I'm not 100%) and this would exclude them from updates, wouldn't it? Aside from some clean up, what are the advantages?  

I'll defer to @hassox on this one

hassox:
It seems like there might be quite a few people still on 1.1. I've asked ericmj for the stats but he needs to go to the CDN logs to find out. I'll come back to this when I have more information.

doomspork:
Very resourceful, thanks @hassox.  :+1:

hassox:
Eric got back to me. The stats for 26 Jan 2016 are here: https://gist.github.com/ericmj/71a37168e761cb7bbf63

Eric said to ignore the 0.13, that's from one IP that is now banned. Also, he said to bear in mind that lots of these are probably CI environments. 

Given that this is only one days worth, it seems that there is still significant < 1.2 usage going on :(

hassox:
@aaronrenner I don't think we can do this yet after the data that we saw from eric. Thoughts?

aaronrenner:
That's fine. I can submit a separate PR with the changes that don't depend on elixir 1.2.

PR #94, opened 2016-01-25
Add config for travis ci

aaronrenner:
This should allow you to hook up guardian to Travis CI and test each pull request various elixir versions going forward.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: elixir
+elixir:
+  - 1.2.0


doomspork:
We should probably consider adding `sudo: false` for faster builds and including an after block for docs, what do you think?

Mimicking what other Elixir projects have would give us something like this:

```yaml
language: elixir
elixir:
  - 1.0.5
  - 1.1.1
  - 1.2.0
sudo: false
otp_release:
  - 17.3
  - 18.1
after_script:
  - mix deps.get --only docs
  - MIX_ENV=docs mix inch.report
```

@hassox / @scrogson — what do you want to do about Elixir and OTP versions?  Should we move to 1.2 & 18.1 or continue to support 1.1 & 17.3?

aaronrenner:
@doomspork I'd be happy to add `sudo: false` and the `after_script` for inch. I submitted #93 which bumps the Elixir dependency to 1.2 and refactors some nested case statements to use the new `with` syntax. Once it's decided if you want to upgrade to Elixir 1.2 or not, I'll update this pull request with the appropriate Elixir versions.

doomspork:
Oh I just noticed `inch` included in there, my bad!  I just quickly grabbed the config from `Plug`,  I meant `mix hex.docs`.

aaronrenner:
I think `mix hex.docs` uploads the documentation to hexdocs.pm. Do we want this to happen automatically for every pull request/commit? I wonder if that should be more of a manual process when a new version is released.

doomspork:
You're right @aaronrenner, I was just thinking about master.

We can probably continue to push the docs manually but it looks like it is possible to automate with TravisCI:

```elixir
after_success:
  - if [ "$TRAVIS_REPO_SLUG" == "ueberauth/guardian" ] 
      && [ "$TRAVIS_BRANCH" == "master" ] 
      && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
      mix deps.get --only docs && MIX_ENV=docs mix hex.docs
    fi
```

Two references I found:

- [Automatically publish JavaDocs](http://benlimmer.com/2013/12/26/automatically-publish-javadoc-to-gh-pages-with-travis-ci/)
- [Conditional `after_success`](http://stackoverflow.com/questions/29082831/perform-after-success-actions-conditionally)

aaronrenner:
Good find, @doomspork. Let's see what @hassox wants to do.

hassox:
I'd love to setup travis. I don't know squat about setting it up though. I'm going to defer to @doomspork on this one. super +1 from me though.

PR #97, opened 2016-01-26
Add dogma

doomspork:
RFC @hassox & @scrogson — This PR adds dogma to the project in both dev and test.  I went ahead and fixed up all of the instances of trailing new lines and pipe chains started without a bare value.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -22,4 +22,8 @@ use Mix.Config
 # Configuration from the imported file will override the ones defined
 # here (which is why it is important to import them last).
 #
+
 if Mix.env == :test || Mix.env == :dev || Mix.env == :docs, do: import_config "#{Mix.env}.exs"
+
+config :dogma,
+  rule_set: Dogma.RuleSet.All
--- a/config/config.exs
+++ b/config/config.exs
@@ -23,7 +23,9 @@ use Mix.Config
 # here (which is why it is important to import them last).
 #
 
-if Mix.env == :test || Mix.env == :dev || Mix.env == :docs, do: import_config "#{Mix.env}.exs"
+if Mix.env == :test ||
+    Mix.env == :dev ||
+    Mix.env == :docs, do: import_config "#{Mix.env}.exs"
 
 config :dogma,
   rule_set: Dogma.RuleSet.All
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -6,4 +6,3 @@ config :guardian, Guardian,
       verify_issuer: true,
       secret_key: "woiuerojksldkjoierwoiejrlskjdf",
       serializer: Guardian.TestGuardianSerializer
-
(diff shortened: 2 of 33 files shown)


config/config.exs
```diff
@@ -22,4 +22,13 @@ use Mix.Config
...
 #
+
 if Mix.env == :test || Mix.env == :dev || Mix.env == :docs, do: import_config "#{Mix.env}.exs"
+
+
+config :dogma,
+  rule_set: Dogma.RuleSet.All,
+
+  override: %{
+    LineLength => [ max_length: 120 ]
```
doomspork: @scrogson do you want to change this back to 80?

doomspork:
If we merge this before #94 we can add dogma as a check.

hassox:
I'm down for this! are there any outstanding style issues that I should fix as part of this?

doomspork:
@hassox there's still 63 dogma errors, I didn't want to go crazy fixing them all until I made sure everyone was on board.

Most of the remaining issue are with LineLength > 120.  If we stick with 80 there will be _a lot_ more :grinning: .  

There's other small issues I can knock out as part of this PR.

doomspork:
Make that 56 remaining.  I missed a few pipe chains and corrected a few other issues.

The remaining issues:
- Missing Moduledocs
- Line Length > 120
- Comparison to boolean in assertions; `assert Permissions.any?(val, val, :other) == true` can be replaced with `assert Permissions.any?(val, val, :other)`

config/config.exs
doomspork: @hassox once we decide on this, the PR should be good to go.

hassox:
Sounds great. Let me know if you want me to try and tackle them.

doomspork:
Totally!  Let's sort out the line length, merge this, and then we can both work at cleaning it up.

aaronrenner:
@doomspork I'm totally a fan of this change, but I'm finding the 80 character line length is a little rough in test descriptions, especially if you want to pass in arguments from the setup block. What are your thoughts on a test like this?

```elixir
test "doesn't call unauthenticated when theres a session with specific key", %{conn: conn} do
  claims = %{ "aud" => "token", "sub" => "user1" }
  #...
end
```

Should I break the test definition onto 2 lines so it meets the character limit?

```elixir
test "doesn't call unauthenticated when theres a session with specific key", 
  %{conn: conn} do
  claims = %{ "aud" => "token", "sub" => "user1" }
  #...
end
```

I like the 80 character line length for application code, but I'm not sure what the best answer is to do here.

doomspork:
@aaronrenner you're right, the test descriptions can be particularly tough to squeeze in under 80.  Truth be told I typically use 120 so I'm learning too :grinning:  

I don't know if this is ideal but what I did was shorten the desc and use single letter variables:

```elixir
# Before
test "fetches the serialized sub from the token for current_resource", %{socket: socket, claims: claims} do

# After
test "fetch serialized sub for current_resource", %{socket: s, claims: c} do
```

In your example I might try something like:

```elixir
test "not unauthenticated when session has specific key", %{conn: conn} do
test "doesn't call unauthenticated with session", %{conn: conn} do

```
Let's see what @scrogson and @hassox have to say.

PR #98, opened 2016-01-27
Fix Guardian.Plug.ErrorHandler

OSChris:
- As of Plug v1.1.0 the Plug.Conn.get_req_header/2 function has a guard clause that checks is_binary on the second argument(key), which fails in the case of passing an atom. After this correction it works like a charm.


msanterre:
Could you add tests for this case?

OSChris:
@msanterre as this was a misuse of the Plug.Conn module I'm not sure how useful a test would be

PR #101, opened 2016-01-29
Fix Guardian.Plug.EnsureAuthenticated so it uses default ErrorHandler

aaronrenner:
According to the documentation, the `Guardian.Plug.EnsureAuthenticated` plug's `:handler` option should default to `Guardian.Plug.EnsureAuthenticated`.  If you currently try to use the plug without the `:handler` option, it raises an error saying "Requires a handler module to be passed".

In addition to updating the default value I also:
* Removed references to old on_failure method
* Refactored tests so options passed into `EnsureAuthenticated.call` are generated by `EnsureAuthenticated.init` method instead of being hard coded at the top of the file.
* Added tests for the `init` method

The only thing that I see as a potential issue is when you run the tests, one of Guardian' deprecation warnings is printed out. I wrote a test to ensure the `:on_failure` argument continues to work even though it's deprecated. Any tips on how to not have the deprecation warning go into the test output would be greatly appreciated.

### DIFF
--- a/lib/guardian/plug/ensure_authenticated.ex
+++ b/lib/guardian/plug/ensure_authenticated.ex
@@ -3,12 +3,12 @@ defmodule Guardian.Plug.EnsureAuthenticated do
   This plug ensures that a valid JWT was provided and has been
   verified on the request.
 
-  If one is not found, the on\_failure function is invoked with the
-  `Plug.Conn.t` object and it's params.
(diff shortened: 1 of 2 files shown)


hassox:
This is awesome :D I just saw that missing default the other day too, great timing!

aaronrenner:
@hassox Just rebased, and the build is green.

hassox:
:metal:

PR #102, opened 2016-01-31
Allow using custom secrets to encode / decode tokens

hassox:
Definitely an RFC.


doomspork:
This looks good to me :+1:

PR #103, opened 2016-01-31
Fix warnings

doomspork:
@hassox this fixes the warnings on master so that the Travis build passes.

### DIFF
--- a/lib/guardian/channel.ex
+++ b/lib/guardian/channel.ex
@@ -37,14 +37,13 @@ defmodule Guardian.Channel do
   directly and authenticating the connection rather than the channel.
   """
   defmacro __using__(opts) do
-    opts = Enum.into(opts, %{})
-    key = Map.get(opts, :key, :default)
+    key = Keyword.get(opts, :key, :default)
 
     quote do
       import Guardian.Phoenix.Socket
 
       def join(room, auth = %{"guardian_token" => jwt}, socket) do
-        case sign_in(socket, jwt, params, key: key) do
+        case sign_in(socket, jwt, params, key: unquote(key)) do
           {:ok, authed_socket, guardian_params} ->
             join(room, Map.merge(params, guardian_params), authed_socket)
           {:error, reason} -> handle_guardian_auth_failure(reason)
--- a/lib/guardian/claims.ex
+++ b/lib/guardian/claims.ex
@@ -110,7 +110,7 @@ defmodule Guardian.Claims do
         Map.put(claims, "exp", iat + years * 365 * 24 * 60 * 60)
       {iat, {years, :year}} ->
         Map.put(claims, "exp", iat + years * 365 * 24 * 60 * 60)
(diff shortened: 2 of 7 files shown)


lib/guardian/permissions.ex
```diff
@@ -170,9 +170,6 @@ defmodule Guardian.Permissions do
     Map.get(c, type, Map.get(c, to_string(type), 0))
   end
 
-  @doc false
```
doomspork: Warning: 
```shell
lib/guardian/permissions.ex:174: warning: this clause cannot match because a 
previous clause at line 168 always matches
```

lib/guardian/plug/verify_header.ex
```diff
@@ -35,7 +35,6 @@ defmodule Guardian.Plug.VerifyHeader do
       # will take the first auth header
       # Authorization: <jwt>
   """
-  import Guardian.Keys
```
doomspork: Warning: 
```shell
lib/guardian/plug/verify_header.ex:38: warning: unused import Guardian.Keys
```

lib/guardian/claims.ex
```diff
@@ -110,7 +110,7 @@ defmodule Guardian.Claims do
         Map.put(claims, "exp", iat + years * 365 * 24 * 60 * 60)
       {iat, {years, :year}} ->
         Map.put(claims, "exp", iat + years * 365 * 24 * 60 * 60)
-      {iat, {_, units}} -> raise "Unknown Units: #{units}"
+      {_iat, {_, units}} -> raise "Unknown Units: #{units}"
```
doomspork: Warning: 
```shell
lib/guardian/claims.ex:113: warning: variable iat is unused
```

lib/guardian/channel.ex
```diff
@@ -37,14 +37,13 @@ defmodule Guardian.Channel do
   directly and authenticating the connection rather than the channel.
   """
   defmacro __using__(opts) do
-    opts = Enum.into(opts, %{})
-    key = Map.get(opts, :key, :default)
+    key = Keyword.get(opts, :key, :default)
```
doomspork: Warning:

```shell
lib/guardian/channel.ex:41: warning: variable key is unused
```

@hassox this was probably a bug, key wasn't being unquoted.  Can you review this change?

aaronrenner:
Looks great to me. :+1: This will be super helpful so my tests will pass.

lib/guardian/channel.ex
hassox: Why the change to Keyword from map? I think this will impact other areas of the code.

hassox: Nm

doomspork: @hassox I made this change because, at first glance, it didn't look like `opts` was being used other than retrieving `key` and converting to a map seemed like overkill.

PR #104, opened 2016-01-31
Update mix.lock to reflect the updated jose dependency

aaronrenner:
`Mix.lock` also needs to be updated for jose 1.6.0.


aaronrenner:
I'll rebase once #103 is merged so the build passes.

hassox:
:+1:

aaronrenner:
@hassox Rebased and green.

PR #105, opened 2016-02-05
Fixed error in ErrorHandler when accept header isn’t set

aaronrenner:
The current version of `Guardian.Plug.ErrorHandler` raises the following error when no `Accept` header is set.

```
** (ArgumentError) argument error
stacktrace:
  :erlang.hd([])
  (guardian) lib/guardian/plug/error_handler.ex:49: Guardian.Plug.ErrorHandler.accept_type/1
  (guardian) lib/guardian/plug/error_handler.ex:12: Guardian.Plug.ErrorHandler.unauthenticated/2
```

I was running into this problem when writing controller tests and I hadn't explicitly set the `Accept` header. This change also adds test coverage to Plug.Guardian.ErrorHandler module.

The build is failing because of the 80 char line length restriction added in #97. Would you be up to changing that restriction to 120 characters so we can maintain more detailed test descriptions and pass in parameters from the setup block? It's kind of a bummer you can't just say 120 characters for only test descriptions, but it doesn't look like dogma allows that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Master (Unreleased)
+
+* Fix error in Guardian.Plug.ErrorHandler when Accept header is unset.
+
 # v 0.10.0
 
 * Add a Guardian.Phoenix.Socket module and refactor Guardian.Channel
--- a/test/guardian/plug/error_handler_test.exs
+++ b/test/guardian/plug/error_handler_test.exs
@@ -1,0 +1,100 @@
+defmodule Guardian.Plug.ErrorHandlerTest do
+  use ExUnit.Case, async: true
+  use Plug.Test
+
+  alias Guardian.Plug.ErrorHandler
+
+  setup do
+    conn = conn(:get, "/foo")
+    {:ok, %{conn: conn}}
+  end
+
+  test "unauthenticated/2 sends a 401 response when text/html", %{conn: conn} do
+    conn = put_req_header(conn, "accept", "text/html")
+
+    {status, headers, body} =
+      conn
+      |> ErrorHandler.unauthenticated(%{})
+      |> sent_resp
+
+    assert status == 401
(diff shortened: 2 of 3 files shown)


hassox:
Nice catch! 

Two quick things. Can you run dogma to remove the dogma warnings.

    mix dogma

and could you also please update the Changelog 

Thanks for all the great PRs!

doomspork:
@aaronrenner we're going to stick with 80 for the time being and see if we can't make it work.  I can help you come up with shortened descriptions if you'd like.

test/guardian/plug/error_handler_test.exs
```diff
@@ -0,0 +1,100 @@
...
+  use Plug.Test
+
+  alias Guardian.Plug.ErrorHandler
+
+  setup do
+    conn = conn(:get, "/foo")
+    {:ok, %{conn: conn}}
+  end
+
+  test "unauthenticated/2 sends a 401 response when accept header is text/html", %{conn: conn} do
```
doomspork: These could be shortened like such: 

```elixir
test "sends a 401 response when text/html", %{conn: conn} do
...
end

test "send a 401 response when application/json", %{conn: conn} do
...
end
```

aaronrenner:
@hassox @doomspork Just shortened the test descriptions and updated the changelog.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Master (Unreleased)
```
doomspork: @hassox what's your plan with the changelog, did you want this to be `v0.10.1`?

hassox: I'm not very experienced with it. What do you think is best?

hassox:
This all LGTM. I'm going to merge. Happy to have the changelog be 'Master' for now. We can always change it if we want later.

Issue #106, opened 2016-02-07
Guardian.Plug.EnsureNotAuthenticated


victorlcampos:
Hi,
I'm coding a login/sign up views that I have guarantee that the user are not Authenticated, if it is, I have to do something(handler).

Could I code this plug and submit a pull request?

doomspork:
I'm :+1: this but let's see what @hassox thinks.

hassox:
Good idea. This seems like a great add. What would you do in the case where someone is logged in?

victorlcampos:
Call a handler ?

doomspork:
I imagine it would work just like `Guardian.Plug.EnsureAuthenticated` only the handler function would be something like `authenticated/2`.

```elixir
# some_controller.ex
alias Guardian.Plug.EnsureNotAuthenticated
plug EnsureNotAuthenticated, [handler: MyApp.MyAuthErrorHandler] when action in [:signin, :signup]

# my_auth_error_handler.ex
defmodule MyApp.MyAuthErrorHandler  do
  def unauthenticated(conn, _params) do
  end

  def authenticated(conn, _params) do
  end
end
```

victorlcampos:
exactly

hassox:
:+1: sounds good. are you comfortable putting up a PR @victorlcampos?

victorlcampos:
I will try, tomorrow, are you ok with this date?

doomspork:
Sounds good @victorlcampos :+1:

victorlcampos:
@doomspork and @hassox PR opened #109

hassox:
This has been merged. Thanks @victorlcampos!

victorlcampos:
Thank you for this anwsome lib

PR #109, opened 2016-02-11
Adding ensure_not_authenticated plug

victorlcampos:
Hi guys,
I come home early from my business trip, so I start the pull request today.

This pull request add a ensure_not_authenticate plug for the case of page that user needs to be  not authenticated like sign_up/sign_in pages.

#106

### DIFF
--- a/lib/guardian/plug/ensure_not_authenticated.ex
+++ b/lib/guardian/plug/ensure_not_authenticated.ex
@@ -1,6 +1,6 @@
 defmodule Guardian.Plug.EnsureNotAuthenticated do
   @moduledoc """
-  This plug ensures that a valid JWT was provided and has been
+  This plug ensures that a invalid JWT was provided and has been
   verified on the request.
 
   If one is not found, the `authenticated/2` function is invoked with the
@@ -32,7 +32,7 @@ defmodule Guardian.Plug.EnsureNotAuthenticated do
       handler: handler,
       key: Map.get(opts, :key, :default),
       claims: Guardian.Utils.stringify_keys(claims_to_check)
-    }
+  }
   end
 
   @doc false
@@ -40,8 +40,8 @@ defmodule Guardian.Plug.EnsureNotAuthenticated do
     key = Map.get(opts, :key, :default)
 
     case Guardian.Plug.claims(conn, key) do
-      { :ok, claims } -> conn |> check_claims(opts, claims)
-      { :error, reason } -> conn
+      {:ok, claims} -> conn |> check_claims(opts, claims)
+      {:error, reason} -> conn
       _ -> conn
     end
   end
@@ -54,16 +54,16 @@ defmodule Guardian.Plug.EnsureNotAuthenticated do
     apply(
       mod,
       meth,
-      [the_connection, Map.merge(the_connection.params, %{ reason: reason })]
+      [the_connection, Map.merge(the_connection.params, %{reason: reason})]
     )
   end
 
-  defp check_claims(conn, opts = %{ claims: claims_to_check }, claims) do
+  defp check_claims(conn, opts = %{claims: claims_to_check}, claims) do
     claims_match = claims_to_check
                    |> Map.keys
                    |> Enum.all?(&(claims_to_check[&1] == claims[&1]))
     if claims_match do
-      handle_error(conn, { :error, :claims_match }, opts)
+      handle_error(conn, {:error, :claims_match}, opts)
     else
       conn
     end
--- a/lib/guardian/plug/ensure_not_authenticated.ex
+++ b/lib/guardian/plug/ensure_not_authenticated.ex
@@ -9,12 +9,12 @@ defmodule Guardian.Plug.EnsureNotAuthenticated do
   ## Example
 
       # Will call the authenticated/2 function on your handler
-      plug Guardian.Plug.EnsureAuthenticated, handler: SomeModule
+      plug Guardian.Plug.EnsureNotAuthenticated, handler: SomeModule
 
       # look in the :secret location.  You can also do simple claim checks:
-      plug Guardian.Plug.EnsureAuthenticated, handler: SomeModule, key: :secret
+      plug Guardian.Plug.EnsureNotAuthenticated, handler: SomeModule, key: :secret
 
-      plug Guardian.Plug.EnsureAuthenticated, handler: SomeModule, aud: "token"
+      plug Guardian.Plug.EnsureNotAuthenticated, handler: SomeModule, aud: "token"
 
   If the handler option is not passed, `Guardian.Plug.ErrorHandler` will provide
   the default behavior.
--- a/lib/guardian/plug/ensure_not_authenticated.ex
+++ b/lib/guardian/plug/ensure_not_authenticated.ex
@@ -3,7 +3,7 @@ defmodule Guardian.Plug.EnsureNotAuthenticated do
   This plug ensures that a invalid JWT was provided and has been
   verified on the request.
 
-  If one is not found, the `authenticated/2` function is invoked with the
+  If one is found, the `authenticated/2` function is invoked with the
   `Plug.Conn.t` object and its params.
 
   ## Example
--- a/lib/guardian/plug/ensure_not_authenticated.ex
+++ b/lib/guardian/plug/ensure_not_authenticated.ex
@@ -27,12 +27,12 @@ defmodule Guardian.Plug.EnsureNotAuthenticated do
     opts = Enum.into(opts, %{})
     handler = build_handler_tuple(opts)
 
-    claims_to_check = Map.drop(opts, [:on_failure, :key, :handler])
+    claims_to_check = Map.drop(opts, [:key, :handler])
     %{
       handler: handler,
       key: Map.get(opts, :key, :default),
       claims: Guardian.Utils.stringify_keys(claims_to_check)
-  }
+    }
   end
 
   @doc false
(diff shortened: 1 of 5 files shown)


victorlcampos:
My English is not so good, I will try to write doc tomorrow, but could you help me with this?
And Could you review my code? I'm very new in Elixir. =-)

Thx for awesome project

lib/guardian/plug/ensure_not_authenticated.ex
```diff
@@ -0,0 +1,85 @@
...
+    }
+  end
+
+  @doc false
+  def call(conn, opts) do
+    key = Map.get(opts, :key, :default)
+
+    case Guardian.Plug.claims(conn, key) do
+      { :ok, claims } -> conn |> check_claims(opts, claims)
+      { :error, reason } -> conn
```
doomspork: This should be `{:error, _reason}`.  

Could you also remove the padding in the tuples? `{:ok, claims}` vs `{ :ok, claims }`

lib/guardian/plug/ensure_not_authenticated.ex
```diff
@@ -0,0 +1,85 @@
...
+
+    {mod, meth} = Map.get(opts, :handler)
+    apply(
+      mod,
+      meth,
+      [the_connection, Map.merge(the_connection.params, %{ reason: reason })]
+    )
+  end
+
+  defp check_claims(conn, opts = %{ claims: claims_to_check }, claims) do
```
doomspork: Please remove the extra padding in this map.

lib/guardian/plug/ensure_not_authenticated.ex
```diff
@@ -0,0 +1,85 @@
+defmodule Guardian.Plug.EnsureNotAuthenticated do
+  @moduledoc """
+  This plug ensures that a valid JWT was provided and has been
```
doomspork: This plug ensures there _isn't_ a valid JWT, right? :grinning:

lib/guardian/plug/ensure_not_authenticated.ex
```diff
@@ -0,0 +1,85 @@
...
+
+  @doc false
+  defp handle_error(conn, reason, opts) do
+    the_connection = conn |> assign(:guardian_failure, reason) |> halt
+
+    {mod, meth} = Map.get(opts, :handler)
+    apply(
+      mod,
+      meth,
+      [the_connection, Map.merge(the_connection.params, %{ reason: reason })]
```
doomspork: Please remove extra padding.

doomspork:
Hi @victorlcampos, this is great thank you!  

For tuples and maps please do not use extra padding: `{:ok, result} vs `{ :ok, result }`.  I commented on a couple instances, could you look through your changes for others?

victorlcampos:
Yes, I'm going to work now, when I come back i finish this. Thx a lot for code review

lib/guardian/plug/ensure_not_authenticated.ex
```diff
@@ -0,0 +1,85 @@
...
+
+  ## Example
+
+      # Will call the authenticated/2 function on your handler
+      plug Guardian.Plug.EnsureAuthenticated, handler: SomeModule
+
+      # look in the :secret location.  You can also do simple claim checks:
+      plug Guardian.Plug.EnsureAuthenticated, handler: SomeModule, key: :secret
+
+      plug Guardian.Plug.EnsureAuthenticated, handler: SomeModule, aud: "token"
```
aaronrenner: These examples should probably mention `Guardian.Plug.EnsureNotAuthenticated`

lib/guardian/plug/ensure_not_authenticated.ex
```diff
@@ -0,0 +1,85 @@
...
+
+  If the handler option is not passed, `Guardian.Plug.ErrorHandler` will provide
+  the default behavior.
+  """
```
aaronrenner: @hassox @doomspork Do we want to default the error handler to the same error handler that `Guardian.Plug.EnsureAuthenticated` uses? It seems like it might get confusing using the same error handler handling when the user is unauthenticated and authenticated. Also, if we go this route, `Guardian.Plug.ErrorHandler` will need a new `authenticated/2` method.

test/guardian/plug/ensure_not_authenticated_test.exs
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  test "calls handler's unauthenticated/2 with no session for default key" do
+    conn = conn(:get, "/foo")
+    opts = EnsureNotAuthenticated.init(handler: TestHandler)
+    ensured_conn = EnsureNotAuthenticated.call(conn, opts)
+    refute must_authenticate?(ensured_conn)
+  end
+
+  test "calls handler's unauthenticated/2 with no session for specific key" do
```
aaronrenner: I think you might have meant `"calls handler's authenticated/2 with no session for specific key"`? There are a couple more references to `unauthenticated` that need to be changed to `authenticated`.

victorlcampos:
@aaronrenner, thx for reply...
Yes, I still need change doc(https://github.com/ueberauth/guardian/pull/109#issuecomment-182661234), about handler, I agree with @doomspork at this comment(https://github.com/ueberauth/guardian/issues/106#issuecomment-181179337)

lib/guardian/plug/ensure_not_authenticated.ex
aaronrenner: What about if the callback was `already_authenticated/2`? Would that might make more sense as a callback in the error handler. Also, what would the default `Guardian.Plug.ErrorHandler` do when this callback was called?

test/guardian/plug/ensure_not_authenticated_test.exs
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  test "it halts the connection" do
+    conn = conn(:get, "/foo")
+    opts = EnsureNotAuthenticated.init(handler: TestHandler, key: :secret)
+    ensured_conn = EnsureNotAuthenticated.call(conn, opts)
+    refute ensured_conn.halted
+  end
+
+  defp must_authenticate?(conn) do
```
aaronrenner: Since this module ensures the user is not authenticated, what would you think about renaming this helper method to `already_authenticated?` Then your assertions would read like

```elixir
assert already_authenticated?(conn)
```

lib/guardian/plug/ensure_not_authenticated.ex
```diff
@@ -0,0 +1,85 @@
...
+      handle_error(conn, {:error, :claims_match}, opts)
+    else
+      conn
+    end
+  end
+
+  defp build_handler_tuple(%{handler: mod}) do
+    {mod, :authenticated}
+  end
+  defp build_handler_tuple(%{on_failure: {mod, func}}) do
```
aaronrenner: This is a new plug so we don't have to handle the deprecated `on_failure:` option.

lib/guardian/plug/ensure_not_authenticated.ex
hassox: `already_authenticated` sounds like a pretty reasonable thing to call this I think. I don't know what the default would do though. maybe redirect to "/" ? That could possibly cause an infinite redirect though if "/" was EnsureNotAuthenticated. Perhaps a required config option for redirection?

lib/guardian/plug/ensure_not_authenticated.ex
doomspork: Good catch!

hassox: :+1:

victorlcampos:
Hi guys,
sory for did not fix this gaps yesterday. I worked 12 hours and when I went home I was very very tired.

test/guardian/plug/ensure_not_authenticated_test.exs
victorlcampos: (y)

victorlcampos:
Hi guys,
I did some changes. What do you think?

test/guardian/plug/ensure_not_authenticated_test.exs
```diff
@@ -0,0 +1,95 @@
...
+
+    assert options == %{
+      claims: %{},
+      handler: {Guardian.Plug.ErrorHandler, :authenticated},
+      key: :default
+    }
+  end
+
+  test "it validates claims and calls through if the claims are not ok" do
+    claims = %{ "aud" => "token", "sub" => "user1" }
```
doomspork: Hey @victorlcampos, for consistency sake could you remove the extra whitespace in your Maps?  It looks like it's this line and lines 46 and 54.

`%{ "aud" => "token", "sub" => "user1" }` -> `%{"aud" => "token", "sub" => "user1"}`

Thanks :+1:

aaronrenner:
@victorlcampos Thanks for the great work! Here are a couple of comments I have.

1. What do you think about changing the callback on the handler to `already_authenticated` instead of just `authenticated`? I am thinking if I'm looking at an error handler module, I'd understand why `unauthenticated` and `unauthorized` are errors, but I would wonder why `authenticated` is there. `already_authenticated` seems like it might give the developer an idea that maybe the user wasn't supposed to be authenticated for a given route. Also, if you have a better name than `already_authenticated` that would be great too.
2. Would you mind adding a simple test to ensure everything works when the developer doesn't pass a `handler:` option? It looks like things will break at the moment because `Guardian.Plug.ErrorHandler` doesn't implement the expected callback. Maybe @hassox or @doomspork can give their two cents on if `Guardian.Plug.ErrorHandler` should provide some sort of default behavior, or if the `handler:` option should just be required.

Thanks for all of the work you've done so far.

victorlcampos:
Hi @aaronrenner,
for now I just halt the conn as default behavior.

aaronrenner:
@victorlcampos Nice work. Looks good to me. :+1:

hassox:
LGTM. Only one thing left. Please add an entry to the changelog that explains the change.

Once that's in, lets merge :D

victorlcampos:
@hassox, could you help me with English?

hassox:
Sure, something like:

    * Adds an EnsureNotAuthenticated plug to handle times when you want to make sure someone is _not_ authenticated. See Guardian.Plug.EnsureNotAuthenticated

hassox:
What you have is fine :D

victorlcampos:
Thx.

This weekend I will add this to gen_guardian generator

Issue #90, opened 2016-01-20
Allow quick sign in for acceptance tests


aaronrenner:
When I'm working in Ruby, I really enjoy how Clearance's [Backdoor Middleware](https://github.com/thoughtbot/clearance#fast-feature-specs) allows me to skip the sign in process when writing browser-level acceptance tests. When in my test environment, I can append the `as` query string parameter when I make a request, and I'm automatically signed in as that user. (Here's how I'd make the request using [Hound](https://github.com/HashNuke/hound)).

``` elixir
navigate_to "/?as=#{user.id}"
```

What would you think about guardian having a middleware that allows developers to be able to skip the sign in process during acceptance tests? To enable this, I was thinking they might add the following to their `web/router.ex` file.

``` elixir
pipeline :browser_session do
  if Mix.env == :test do
    plug Guardian.Plug.Backdoor
  end

  plug Guardian.Plug.VerifySession
  plug Guardian.Plug.LoadResource
end
```

I'd be happy to work on this and I'm still thinking over the implementation details, but I wanted to run it by you before I start.

hassox:
Putting the value into the session is a bit of a pain. A backdoor would be awesome :D

aaronrenner:
I've got a proof of concept done for this and will be submitting a PR in the next couple of days.

hassox:
cool

aphillipo:
I'm looking for something like this too.  Not sure of the correct way to do this without the Backdoor Middleware... What do people do in the mean time?

hassox:
@aphillipo Down the bottom of http://blog.overstuffedgorilla.com/simple-guardian/ there's a way to work with session stored tokens. The backdoor middleware will make this dance much simpler.

jamesvl:
The example in the blog mentioned above is what I was looking for - would you consider adding that snippet to the docs or readme file?

And although it's slower, I like that method better than the backdoor middleware which uses GET params to setup a fake user - for purposes of testing various routes, you are now no longer testing the exact route (since it has a query string), and there's a chance it conflicts with expected parameters (unlikely, but possible).

doomspork:
@aaronrenner / @hassox — any updates on this functionality and PR #120?

scrogson:
Tracking this in https://github.com/ueberauth/guardian/pull/354

PR #110, opened 2016-02-11
Refactor Plug Tests

aaronrenner:
I'd love to get your feedback on this PR. When dogma was introduced, many of the plug tests were refactored to have lines like this:

```elixir
:get |> conn("/foo") |> Guardian.Plug.set_claims({ :ok, claims })
```
This takes a minute for me to figure out what's going on and distracts me from the core things that are happening in the test. This PR moves making that arbitrary GET request into the setup block, so each test is passed a connection that is already ready to go.

```elixir
setup do
  conn = conn(:get, "/foo")
  {:ok, %{conn: conn}}
end

test "validates claims and calls through if claims are ok", %{conn: conn} do
  # conn is already setup now...
end
```

The other thing that's repeated over and over is this:

```elixir
opts = EnsureAuthenticated.init(handler: TestHandler, aud: "token")
ensured_conn = EnsureAuthenticated.call(conn, opts)
```
This code is just boilerplate for running the plug. It will always run the `init` function with certain arguments, and then pass the generated options into the `call` function as the 2nd parameter. I figured this boilerplate doesn't add to the readability of the test, so I extracted it into a `run_plug` helper. This allows a whole test to be written like this:

``` elixir
test "validates claims and calls through if claims are ok", %{conn: conn} do
  claims = %{ "aud" => "token", "sub" => "user1" }
  
  ensured_conn =
    conn
    |> Guardian.Plug.set_claims({ :ok, claims })
    |> run_plug(EnsureAuthenticated, handler: TestHandler, aud: "token")
  
  refute must_authenticate?(ensured_conn)
end
```

I'd love to hear your feedback on these changes. Do you think it makes the tests easier to read, or does it hide too much? If you like it, I'd be happy to update the rest of the plug tests before you merge this PR.

### DIFF
--- a/test/guardian/plug/ensure_permissions_test.exs
+++ b/test/guardian/plug/ensure_permissions_test.exs
@@ -1,6 +1,7 @@
 defmodule Guardian.Plug.EnsurePermissionTest do
   use ExUnit.Case, async: true
   use Plug.Test
+  import Guardian.TestHelper
 
   alias Guardian.Plug.EnsurePermissions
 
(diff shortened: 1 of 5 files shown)


doomspork:
This is great @aaronrenner, I think it cleans the tests up quite nicely.

aaronrenner:
Glad you like it, @doomspork. I'll get to work on the rest of the tests.

aaronrenner:
@hassox and @doomspork,  I just pushed my refactoring work for the rest of the plug tests. I also added a 2nd commit where I created Guardian.PlugCase in test/support to share the new `run_plug` function and automatically use `Plug.Test`. It's very similar to `MyApp.ConnCase` in a Phoenix app and maybe I should have named it `Guardian.ConnCase`.

Let me know what you think. If you like it, I can squash the changes and have it merged in a single commit.

hassox:
<3 this is great!

aaronrenner:
After a little more thinking, I'm not quite sold on the `PlugCase` code I wrote. Let's hold off on merging this for right now.

doomspork:
@aaronrenner out of curiosity, is that a result of the feedback you got on Plug?

aaronrenner:
@doomspork How did you know about that PR? Actually, I want to tackle #90 in my next PR and write an integration test that runs the following plugs

```elixir
Guardian.Plug.Backdoor
Guardian.Plug.VerifySession
Guardian.Plug.LoadResource
```
 to make sure that the backdoor sign in flow works correctly. I am still a fan of the `run_plug` helper, regardless of the feedback to the PR. I'm thinking maybe I should have put that helper in a separate module that's imported into the test case, instead of combining that function and additional imports into a single `Guardian.PlugCase`. Also, if I'm writing an integration test, it might be weird to have to add `use Guardian.PlugCase` in my test just to use that helper. I think this might have been a premature refactoring on my part.

I would be curious to hear your feedback on the `run_plug` helper, as well as `Guardian.PlugCase`. I'll try to get to finishing off this PR later today.

doomspork:
I'm all seeing @aaronrenner, obviously! :stuck_out_tongue_closed_eyes: 

There's a few repos I keep up-to-date on and Plug is one of them.  When you popped up I popped in to see what your changes were. :neckbeard: 

Regarding the feedback, I think within the scope of Plug it was warranted but doesn't necessarily apply here.  It makes sense for Plug not to obfuscate the lifecycle or misrepresent the flow in tests but for us the `init/1` and `call/2` lifecycle isn't as relevant.

That's my $0.02 :moneybag:

aaronrenner:
@doomspork @hassox I think this is ready to merge.

test/guardian/plug/ensure_permissions_test.exs
```diff
@@ -12,59 +13,54 @@ defmodule Guardian.Plug.EnsurePermissionTest do
...
     claims = %{ "pem" => %{ "default" => pems } }
 
-    expected_conn = :get
-                    |> conn("/get")
-                    |> Guardian.Plug.set_claims({ :ok, claims })
-                    |> Plug.Conn.fetch_query_params
-                    |> EnsurePermissions.call(opts)
+    expected_conn =
+      conn
+      |> Guardian.Plug.set_claims({ :ok, claims })
```
doomspork: @aaronrenner could you remove the whitespace padding in tuples and maps as you change them?

aaronrenner: Just fixed the whitespace padding for tuples on the lines I changed. Do you want me to go through and change all of them in these files?

doomspork: @aaronrenner no, I don't think that's necessary.  We can update them as we go through and change code.

doomspork:
LGTM :+1: 

@hassox any additional thoughts?

hassox:
LGTM

PR #112, opened 2016-02-13
Rename file claim_valiation to claim_validation

KushalP:
This is small typo fix that was erroneously added in f981bbe.


doomspork:
Thanks @KushalP :+1:

PR #115, opened 2016-02-17
Update to use credo rather than dogma

hassox:
Also updates poison.

Credo seems to pick up a _lot_ more than dogma and that's for sure something I need.

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -35,9 +35,9 @@ defmodule Guardian do
   The resource will be run through the configured serializer
   to obtain a value suitable for storage inside a JWT.
   """
-  @spec encode_and_sign(any) :: { :ok, String.t, Map } |
-                                { :error, atom } |
-                                { :error, String.t }
+  @spec encode_and_sign(any) :: {:ok, String.t, Map} |
+                                {:error, atom} |
+                                {:error, String.t}
   def encode_and_sign(object), do: encode_and_sign(object, nil, %{})
 
   @doc """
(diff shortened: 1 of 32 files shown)


lib/guardian.ex
```diff
@@ -35,9 +35,9 @@ defmodule Guardian do
   The resource will be run through the configured serializer
   to obtain a value suitable for storage inside a JWT.
   """
-  @spec encode_and_sign(any) :: { :ok, String.t, Map } |
```
doomspork: Credo catches these?  Awesome!

hassox: there were so many :|

doomspork:
Looks good to me @hassox :+1:

aaronrenner:
Good find @hassox. It looks like the [default config](https://github.com/rrrene/credo/blob/master/.credo.exs) doesn't lint any of the tests. I'm fine with not linting the tests, but I just wanted to point that out.

hassox:
@aaronrenner do you happen to know the Magic Config to get it to lint the tests? I think we should probably include those

aaronrenner:
It looks like you create a `.credo.exs` file in either `config/` or the root folder, and then change the `files:` > `included:` array. https://github.com/rrrene/credo/blob/master/.credo.exs#L20. I'm not sure if you have to copy the entire config file into your project or if it merges with the default settings. I'd be happy to take a look at it in the morning if you want, but right now I'm off to bed.

hassox:
Thanks for the pointer. I'll pick it up on the train into work in the am

PR #120, opened 2016-02-22
Added Guardian.Plug.Backdoor

### LINKED ISSUE #90: Allow quick sign in for acceptance tests

aaronrenner:
This plug helps developers write faster acceptance tests, by not requiring to go through the entire sign in process for each test. This feature is inspired by Clearance’s backdoor: https://github.com/thoughtbot/clearance#fast-feature-specs

Any feedback is welcome and greatly appreciated.

Fixes #90.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Master (Unreleased)
 
+* Add Guardian.Plug.Backdoor for acceptance testing
+
 # v 0.10.1
 
 * Fix error in Guardian.Plug.ErrorHandler when Accept header is unset.
--- a/lib/guardian/plug/backdoor.ex
+++ b/lib/guardian/plug/backdoor.ex
@@ -1,0 +1,115 @@
+defmodule Guardian.Plug.Backdoor do
+  @moduledoc """
+  This plug allows you to bypass authentication in acceptance tests
+  by passing the token needed to load the current resource directly to your
+  serializer, via a query string parameter.
+
+  ## Installation
+
+  Add the following to your Phoenix router before any other Guardian plugs.
+
+  ```
+  if Mix.env == :test do
+    plug Guardian.Plug.Backdoor
+  end
+  ```
+
+  ***This plug is designed for acceptance testing scenarios and should
+  never be added in a production environment.***
+
+  ## Usage
+
+  Now that `Guardian.Plug.Backdoor` is installed, it's time to log in.
+  If you're using [Hound][hound], you can write the following code.
+
+  ```
+  navigate_to "/?as=User:5"
+  ```
+
+  If you aren't using [Hound][hound], a simple `GET /?as=User:5` request will
+  work.
+
+  When the `Guardian.Plug.Backdoor` plug runs, it passes along the value
+  of the `as` parameter directly to your application's Guardian serializer.
+
+  ```
+  defmodule MyGuardianSerializer do
+    @behaviour Guardian.Serializer
+
+    def from_token("User:" <> user_id) do
+      # Find and return the user object
+      {:ok, %{id: user_id}}
+    end
+    def from_token(_token) do
+      {:error, "Invalid token"}
+    end
+
+    def for_token(user) do
+      # Serialize the user into a single token
+    end
+  end
+  ```
+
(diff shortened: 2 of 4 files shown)


doomspork:
Nice work @aaronrenner :+1:

hassox:
@aaronrenner nice. 

Couple of thoughts. Do you think we should just be passing the sub field in to create a full token? It seems like this would not allow us to include things like permissions or other claims to be included in the bypassed token. 

What would you think of either passing in the full token via the url or header and using that to sign in? The Verify* plugs are really just for locating and verifying a token from _somewhere_. I was thinking that this might be similar. A bypass plug would locate the token from X location and load it into the current connection/session.

Thoughts?

aaronrenner:
@hassox That's a great point that this doesn't allow the token to include permissions or claims. I do like the simplicity of this approach since it doesn't require a new developer to know the all of the details of generating a JWT.

However, I think you're right that it might be a good idea to pass in the token via the URL. That would make the backdoor sign in code look something like this.

```elixir
{:ok, jwt, _} = Guardian.encode_and_sign(user, :token)
navigate_to "/?jwt=#{jwt}"
```

I will have to make sure the backdoor code still calls the `Guardian.Plug.sign_in` function to trigger the callbacks and set the current resource, claims and token. I'll probably decode the token in the Backdoor Plug and then pass along the appropriate params to `sign_in`.

I guess really the only time you'd need this backdoor is if you using the `VerifySession` plug. You wouldn't need this if you are using the `VerifyHeader` plug, since the token is just passed in over the Authorization header.

Let me know if this approach sounds right.

hassox:
I think that sounds reasonable mate. I was a bit torn with putting it in the url but I think it makes sense if someone's using hound or similar.

One of the issues with using `Guardian.Plug.encode_and_sign` in this plug is that it would change the jti, iat and exp which would probably be bad. 

I think this needs to:

1. Decode and verify the token
2. Put all the relevant assings onto the conn object
3. Store the token on the session
4. Remove the jwt from the params

I think if we just use `encode_and_sign` in the bypass plug we'll end up with a different token than the one passed in.

doomspork:
Whatever happened to this @aaronrenner and @hassox?

doomspork:
@aaronrenner any interest in finishing this?

aaronrenner:
@doomspork Yeah, I think it would be good if I finish this up. I'll start looking at it in the evenings again.

doomspork:
Awesome 🎉 

Thank you @aaronrenner 👍

doomspork:
@hassox do you know what is left to finish here?  I can take this over to get it merged in.

avocade:
👍

jayjun:
@aaronrenner Bumping this, because it'll be super handy. Happy to assist. 🙂

aaronrenner:
@jayjun Unfortunately I don't have the time to work on this. Feel free to adapt what I'm submitted if it's useful and submit a PR.

doomspork:
@jayjun I'd be happy to lend a hand if you need any assistance getting this finished up and ready to merge 😀

jayjun:
@doomspork I'll have to grok Guardian before attempting this. If I (ever) get time to revisit tests, I'll take a crack, otherwise I'm currently swamped too.

doomspork:
Closing in favor of #354

PR #123, opened 2016-03-09
fix typespec

sasa1977:
Fixes the incorrect type specification in `Guardian.Serializer`


doomspork:
Thanks @sasa1977 :+1:

PR #127, opened 2016-03-20
Fix refresh to revoke the old token

hassox:
Addresses the bug reported in https://github.com/ueberauth/guardian/pull/121


doomspork:
Looks good @hassox :+1:

PR #132, opened 2016-04-08
use resource from conn if already present

manukall:
I noticed that the `LoadResource` assumed that the resource was inside a `{:ok, resource}` tuple inside the `conn`. `Guardian.Plug.set_current_resource` however just puts the resource into the `conn` without wrapping it in a tuple. This made testing a bit hard. With this fix it's possible to just use a helper like the following in controller tests:
```elixir
def sign_in(conn, user \\ %User{}) do
  conn
    |> Guardian.Plug.set_claims({:ok, %{}})
    |> Guardian.Plug.set_current_resource(user)
end
```

### DIFF
--- a/lib/guardian/plug/load_resource.ex
+++ b/lib/guardian/plug/load_resource.ex
@@ -21,9 +21,7 @@ defmodule Guardian.Plug.LoadResource do
     key = Map.get(opts, :key, :default)
 
     case Guardian.Plug.current_resource(conn, key) do
-      {:ok, _} -> conn
-      {:error, _} -> conn
-      _ ->
+      nil ->
         case Guardian.Plug.claims(conn, key) do
           {:ok, claims} ->
             result = Guardian.serializer.from_token(Map.get(claims, "sub"))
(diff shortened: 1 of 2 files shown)


manukall:
Hey, could I get some input on this. I'm currently depending on my form from my projects, but would like to depend on the hex package again. I thought this was a bugfix, but I'm not completely sure :-)

hassox:
@manukall sorry for the delay. The reason it needs to go into a tuple is because it's completely unknown where it's coming from. Api, memory, db. The tuple provides a way to find out what happened in the case that something when wrong. I think the better option would be to make sure that `set_current_resource` wraps it in a tuple when it puts it in there..

manukall:
@hassox thanks for your reply. i'll reset the `LoadResource` plug and change `set_current_resource`. Would you want `Plug.current_resource` to still return just the resource or `{:ok, resource}` then? (the latter would obviously break the existing api.)

manukall:
ping @hassox

PR #133, opened 2016-04-08
loosen poison requirement

manukall:
this allows poison versions required by the current `phoenix_ecto` version. tests are still green, so maybe poison `~> 2.1` is not actually required yet

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -42,7 +42,7 @@ defmodule Guardian.Mixfile do
 
   defp deps do
     [{:jose, "~> 1.6"},
-     {:poison, "~> 2.1"},
+     {:poison, ">= 1.3.0 and < 3.0.0"},
      {:plug, "~> 1.0"},
      {:ex_doc, "~> 0.10", only: :docs},
      {:earmark, ">= 0.0.0", only: :docs},
--- a/mix.exs
+++ b/mix.exs
@@ -42,7 +42,7 @@ defmodule Guardian.Mixfile do
 
   defp deps do
     [{:jose, "~> 1.6"},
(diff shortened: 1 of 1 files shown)


mix.exs
```diff
@@ -42,7 +42,7 @@ defmodule Guardian.Mixfile do
 
   defp deps do
     [{:jose, "~> 1.6"},
-     {:poison, "~> 2.1"},
+     {:poison, ">= 1.3.0 and < 3.0.0"},
```
doomspork: @hassox do you know which version of Poison we require?

hassox: So long as it's above 2 I'm pretty sure we're good. So long as we can smash JSON we're good. I don't think there is even a real requirement from Guardians side other than it be there since we're just doing string keys. I'd be happy to merge this but maybe we can just say >= 1.3.0?

doomspork: I think `>= 1.3.0` is a good option.

jamesvl:
Any plans to merge this is in, or at least make a new minor release soon?

With Ecto 2.0 in rc stage, much of the Phoenix stack is ready for Poison ~> 2.0, but the 0.10.1 release of Guardian is still back on 1.3.0.

manukall:
i pushed another commit allowing poison >= 1.3.0. is this ok?

jamesvl:
I think what was already in there should have been fine - it's more about the tagged release (0.10.1) being back on 1.5 still. Or would you recommend just setting up mix to pull in from the Git repo head?

doomspork:
Looks good to me @manukall.  @hassox any thoughts or can we merge this?

PR #134, opened 2016-04-11
Fix: change doc to reflect actual behavior for Guardian.Plug.sign_in

hannahhoward:
This was mainly to confirm my own understanding. The docs say passing a type to the Guardian.Plug.sign_in/3 or Guardian.Plug.sign_in/4 functions will set the 'aud' field on the JWT. However, my reading of Guadrian.encode_and_sign suggests it sets the 'typ' field instead.


doomspork:
@hassox any thoughts on this change?

PR #147, opened 2016-04-26
Support for checking one_of perm sets in EnsurePermissions.

ashrafhasson:
To enable authorizing the request when one of the permissions' sets is matched.

Thanks for your work :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,19 @@ defmodule MyApp.MyController do
 end
 ```
 
+When permissions' sets are specified through a `:one_of` map, the token is searched for at least one
+matching permissions set to allow the request. The first set that matches will allow the request.
+If no set matches, the `:unauthorized` function is called.
+
+```elixir
+defmodule MyApp.MyController do
+  use MyApp.Web, :controller
+
+  plug Guardian.Plug.EnsurePermissions, handler: MyApp.MyAuthErrorHandler, 
+    one_of: [%{default: [:read, :write]}, %{other: [:read]}]
+end
+```
+
 ### Pipelines
 
 These plugs can be used to construct pipelines in Phoenix.
--- a/lib/guardian/plug/ensure_permissions.ex
+++ b/lib/guardian/plug/ensure_permissions.ex
@@ -16,6 +16,12 @@ defmodule Guardian.Plug.EnsurePermissions do
                               default: [:profile],
                               handler: SomeMod
 
+      # read AND write permissions for the admin set
+      # OR :profile for the default set
+      plug EnsurePermissions, one_of: [%{admin: [:read, :write]},
+                              %{default: [:profile]}],
+                              handler: SomeMod
+
       # admin :read AND :write for the claims located in the :secret location
       plug EnsurePermissions, key: :secret,
                               admin: [:read, :write],
@@ -36,7 +42,17 @@ defmodule Guardian.Plug.EnsurePermissions do
     on_failure = Map.get(opts, :on_failure)
     key = Map.get(opts, :key, :default)
     handler = Map.get(opts, :handler)
-    perms = Map.drop(opts, [:handler, :on_failure, :key])
+
+    perm_sets = case Map.get(opts, :one_of) do
+      nil ->
+        single_set = Map.drop(opts, [:handler, :on_failure, :key, :one_of])
+        if Enum.empty?(single_set) do
+          []
+        else
+          [single_set]
+        end
+      one_of -> one_of
+    end
 
     if handler do
       handler = {handler, :unauthorized}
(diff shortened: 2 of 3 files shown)


lib/guardian/plug/ensure_permissions.ex
```diff
@@ -36,7 +42,8 @@ defmodule Guardian.Plug.EnsurePermissions do
     on_failure = Map.get(opts, :on_failure)
     key = Map.get(opts, :key, :default)
     handler = Map.get(opts, :handler)
-    perms = Map.drop(opts, [:handler, :on_failure, :key])
+    one_of = Map.get(opts, :one_of)
+    perms = Map.drop(opts, [:handler, :on_failure, :key, :one_of])
```
hassox: I think this one would be better written to have the same shape. Something like

```elixir
single_set = Map.drop(opts, [:handler, :on_failure, :key, :one_of])
perm_sets = Map.get(opts, :one_of, [single_set])
```
Then the keys down below would be:

```elixir
%{
  handler: handler,
  key: key,
  perm_sets: perm_sets
}
```

lib/guardian/plug/ensure_permissions.ex
```diff
@@ -63,19 +71,38 @@ defmodule Guardian.Plug.EnsurePermissions do
...
         perms = Map.get(opts, :perms, %{})
-        result = Enum.all?(Map.get(opts, :perm_keys), fn(perm_key) ->
-          found_perms = Guardian.Permissions.from_claims(claims, perm_key)
-          Guardian.Permissions.all?(
-            found_perms,
-            Map.get(perms, perm_key),
-            perm_key
-          )
-        end)
+        result =
```
hassox: If you have a list of permission sets from the initializer we can just check all the sets. I'm thinking something along the lines of:

```elixir
result = matches_permissions?(conn, key, Map.get(opts, :perm_sets))

if result, do: conn, else: handle_error(conn, opts)
```

```elixir
def matches_permissions?(_, _, []), do: true
def matches_permissions?(conn, key, sets) do
  case Guardian.Plug.claims(conn, key) do
    {:ok, claims} -> 
      Enum.any?(sets, &matches_permission_set?(claims, &1))
    _ -> false
  end
end

defp matches_permission_set?(claims, set) do
  Enum.all?(set, fn({key, required_perms}) -> 
    Guardian.Permissions.from_claims(claims, key)
    |> Guardian.Permissions.all?(required_perms, key)
  end)
end
```

ashrafhasson: I like this approach :) but when does `matches_permissions?(_, _, [])` get its pattern matched? Also, should it not be private?
Another note is below regarding the use of Enum.any.

ashrafhasson:
Thanks for the feedback @hassox really appreciate it.
Hmmm, I thought if we don't specify a `one_of` to the initializer we should still honour allowing multiple sets checked but instead of OR'ing them, they'd be logically AND'ed, or so I thought the default (current) behaviour is. I've tested with multiple permission sets without the `one_of` and it seems as if all sets must match.

ashrafhasson:
I found that [these two test cases](https://github.com/ueberauth/guardian/blob/master/test/guardian/plug/ensure_permissions_test.exs#L84-L124) would fail _if we don't check all permission sets_ when `one_of` is not present.

lib/guardian/plug/ensure_permissions.ex
hassox: You're right it should be private. It would be matched when no sets/permissions were declared. An unlikely case but possible.

hassox:
If we don't specify `one_of` we should still honor current behaviour. I _think_ should do it but I haven't tested.

```elixir
if Map.get(opts, :one_of) do
  perm_sets = Map.get(opts, :one_of)
else
  single_set = Map.drop(opts, [:handler, :on_failure, :key, :one_of])
  if Enum.empty?(single_set) do
    perm_sets = []
  else
    perm_sets = [single_set]
  end
end
```

README.md
```diff
@@ -164,6 +164,19 @@ defmodule MyApp.MyController do
 end
 ```
 
+When permissions' sets are specified through a `:one_of` map, the token is searched for at least one
+matching permissions set to allow the request. The fist set that matches will allow the request.
```
doomspork: "The first set"

ashrafhasson: @doomspork lol, yes and sorry about the typo, I'll fix that when I get to it. Thank you :)

ashrafhasson:
@hassox I've applied your changes to the pull request, please let me know if you'll need a rebase and a squash before accepting this PR.

Thanks,

hassox:
@ashrafhasson looks good mate :D After you rebase, could you please add an entry to the changelog?

Issue #111, opened 2016-02-13
Proposal Guardian Refresh Tokens


Hanspagh:
A common use case for many api's that are consumed by mobile app's is to get a new token without re-authenticating to avoid contentiously prompt users with login screens. This is most commonly done with refresh tokens. Would it be an idea to make a project similar to GuardianDb to implement this behavior?

hassox:
Hey @Hanspagh. Guardian provides a refresh! function. Checkout Guardian.refresh!

Hanspagh:
Yes, I know but that only allows us to make a new token from an valid existing one. The feature I am requesting is to create a new token without having a valid token, but from a refresh token. Like described here https://auth0.com/docs/refresh-token

hassox:
Refresh tokens are tricky. There's a couple of ways to do them that spring to mind.

1. Create a non-expiring token containing/mapping the 'claims' and store it in the db. When issuing a JWT from it, use the claims from the refresh token and encode them into an access token. This requires that you have a database and lookup the refresh token. This method would be outside the scope of Guardian since you'd have to maintain some state on your server.

2. Create a token with Guardian of type 'refresh' that has a long expiry (years) containing all claims required. You can then use this token to exchange for a type of 'access' containing the same claims with a much shorter exp. This does not require you to store the token in a db - although you should probably use something like GuardianDb so that it can be revoked. This would require the addition of an 'exchange' function in guardian so that you could exchange one type of token for another. 

I'd be down for adding an exchange function I think. It's been on my todo list for a while I just haven't had the motivation to actually write it.

Thoughts?

Hanspagh:
My first thought was to implement it like you described in 1. and then add it as dependency like GuardianDb.

But 2. seems to integrate much better into the existing code and being able to reuse GuardianDb instead of writing a db integration again seems to be a big win.
I am still a bit new to Elixir but I would be happy to help with implementing this :)

hassox:
@Hanspagh I'd be happy to review something for #1 but I don't think it should go into the Guardian lib.

Hanspagh:
When you say it should not go in the Guardian lib, do you then mean it should be a lib like GuardianDb or should it be part of the core.

hassox:
I think it should be a separate lib like guardian db

Hanspagh:
Awesome, Will start working on something

hassox:
Sounds great. I'd love to see it when you have something.

Hanspagh:
[...]

PR #150, opened 2016-05-12
add functionality to exchange one token to another

Hanspagh:
Makes it possible to exchange a token with type 'refresh' to a token with type 'token', like we discussed  in #111, it will play nicely with guardianDB once  hassox/guardian_db#15


lib/guardian.ex
```diff
@@ -190,6 +190,50 @@ defmodule Guardian do
     end
   end
 
+
+  @doc """
+  Exchange a token with type 'refresh' for a token with type 'token'
+  with the same claims but a shorter expiry time.
+
+  The token with typ 'refresh' will be revoked after the exhange
```
hassox: Is this the normal proceedure to revoke the refresh token? I thought these stayed valid and could be used multiple times.

hassox:
This looks pretty good, just the q about revoking a refresh token. I could be wrong but my understanding is that they should stick around.

Also can you please add some words to the readme and changelog?

Thanks :D

Hanspagh:
I guess we could make it up to the user to revoke it, or create two different methods exchangeAndRevoke and exchange. What do you think?
I am updating the readme as we speak.

Hanspagh:
I just realized that you can use the refresh token as a normal token. How can we fix this?

lib/guardian.ex
```diff
@@ -366,10 +410,16 @@ defmodule Guardian do
     |> Map.delete("perms")
   end
 
-  defp set_ttl(claims) do
-    claims
-    |> Guardian.Claims.ttl
-    |> Map.delete("ttl")
+  defp set_ttl(claims, type) do
+    if type === "refresh" do
```
doomspork: Do we want to support `:refresh`?  We use atoms elsewhere.

README.md
```diff
@@ -476,7 +478,19 @@ end
...
 Once the new token is created, the old one is revoked before returning the new
-token.
+token. 
+
+### Refresh Tokens
+
+You can create refresh token, which by default are long living. These can later be exhanged for shorter living access tokens, which can be used as a authorization token.
+
+```elixir
+    #issue a long living refresh token
```
doomspork: Could you add a space after `#` and capitalize Issue?  Same feedback for the comment below.

lib/guardian.ex
Hanspagh: yes, but this ensures that we can set a default living time for the refresh tokens

hassox:
@Hanspagh correct you could use it. Once you start using different types of tokens EnsureAuthenticated should check the typ field. I use `access` tokens for normal calls and other types as appropriate.

hassox:
As an example, here's my plug for one of my apps

```elixir
  pipeline :api_auth do
    plug Guardian.Plug.VerifyHeader, realm: "bearer"
    plug Guardian.Plug.EnsureAuthenticated, typ: "access", handler: MyAwesomeApp.Api.V1.ApiAuthErrorHandler
  end

```

Hanspagh:
Makes sense, my only concern is that if someone isn't aware of this, you could authenticate your self with a refresh token, which defeats the whole purpose of this. So we might consider having EnsureAuthenticated check for the typ "token" per default. What do you think?

hassox:
I'm not sure if that would be good or not. I'm thinking about pipelines and setting the token type there. It _might_ be ok. Let me percolate on it a bit. Also I think we should change the default type to 'access' rather than 'token'. Token doesn't really mean anything.

Hanspagh:
I think you are right 'access' makes much more sense then token. If we don't make it a default in EnsureAuthenticated as well,  I think we should at least inform the user about the implications.

Hanspagh:
Any progress on this?

hassox:
Hey @Hanspagh thanks for pinging on this one. I think we definitely need an `exchange` function but I don't think we should limit it to only `refresh` tokens. You could exchange any kind. I do like the refresh TTL of the token  too and would love to see it come in.

I _think_ what would be great is:

1. we add an exchange function that can handle whatever type of token and exchange it for another. This should take the existing token, the new type, the old type and any options to modify
2. We should change the existing default from `token` to access.
3. Update the README to define the implications of using different token types
4. add a configuration entry for `token_ttl` where you can specify the defaults of each type of token.

Thoughts?

Hanspagh:
I think the overall idea sounds really solid.
This allows to define your own token type and still benefit from the exchange functionality. The only downside I see is that someone could potentially exchange a "access" token for a "refresh" token. 
We could predefine these two types and enforce some behavior to prevent that. What do you think?

hassox:
Hmm true facts. I guess there's a couple of ways I can think of that we could deal with that. 

1. Add a config option that provides a mapping from -> to
2. Add into the exchange function signature the expected from -> to. If the from doesn't match it doesn't work

I think 2 seems more natural to me. That way the calling code would be responsible:

```elixir
Guardian.exchange(jwt, "refresh", "access")
```
Or in the case where it's many:

```elixir
Guardian.exchange(jwt, ["refresh", "rememberMe"], "access")
```

Thoughts?

Hanspagh:
Seems good, I will start working on it

Hanspagh:
Here is first shot at the proposed solutions. Still missing docs and being able to modify the token created
Some questions:
* Should we keep the ttl config, so we wont break existing projects, and just use it as a fallback if the token_ttl does not exists.
* We might consider lowering the fallback ttl from 1000.000.000s to something more sense able or throw and error if there is no config

lib/guardian.ex
```diff
@@ -189,6 +189,51 @@ defmodule Guardian do
...
+  end
+
+  defp correct_typ?(_claims, _typ) do
+    false
+  end
+
+  defp do_exchange(from_typ, to_typ, claims) do
+    if correct_typ?(claims, from_typ) do
+      {:ok, resource} = Guardian.serializer.from_token(claims["sub"])
+      case encode_and_sign(resource, to_typ, %{}) do
```
hassox: I think we want to drop some of the claims here and use the remaining claims. If we just use an empty map we'll lose any custom claims that were embedded in the 'refresh' token.

lib/guardian.ex
```diff
@@ -189,6 +189,51 @@ defmodule Guardian do
...
+                              {:error, String.t}
+
+  def exchange(old_jwt, from_typ, to_typ) do
+    case decode_and_verify(old_jwt) do
+      {:ok, found_claims} -> do_exchange(from_typ, to_typ, found_claims)
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  defp correct_typ?(claims, typ) when is_binary(typ) do
```
hassox: I think we need another clause here to deal with the case of an atom.

Hanspagh:
I will have a look. I have a very busy work week coming up, but I will be back in action the 27.

Hanspagh:
I am sorry for the huge delay on this, but I have fixed the two issues you mentioned namely; to handle atom types and not drop all claims when doing the exchange

doomspork:
Thank you @Hanspagh! @hassox any additional feedback?

Hanspagh:
If you don't have anymore feedback, I will update the docs and we can finish this.

Moussenger:
When this pull request will be merged?

doomspork:
@hassox any final feedback?  @Hanspagh do you mind rebasing?

Hanspagh:
I will update the docs and rebase it this weekend

doomspork:
Thank you @Hanspagh

Hanspagh:
Okay, I did the rebase, and added some docs.
One thing we didn't talk about is the ability to have more than one token defined the the the config file.
Where should document this?

Hanspagh:
I am sorry for reviving this again, but I am pretty sure that we could just merge this?

doomspork:
Sorry for the delays @Hanspagh, I've not been following this thread.  Let me review the changes and see about merging.

PR #157, opened 2016-06-01
Call JOSE.JWK from an anonymous function

yoshdog:
Without wrapping it in an anonymous function, it tries to run the `JOSE.JWK` method when the config file gets loaded and it blows up with a: 
```
** (Mix.Config.LoadError) could not load config config/config.exs
    ** (UndefinedFunctionError) undefined function JOSE.JWK.from_file/1 (module JOSE.JWK is not available)

```
error


doomspork:
Thank you @yoshdog 👍

PR #165, opened 2016-06-18
Change default token type from "token" to "access"

hassox:


doomspork:
@hassox this looks good, what's the rationale behind this change?

hassox:
The 'token' type is pretty useless. Whenever we talk about token types (taking a nod from OAuth) we usually talk about access vs refresh tokens etc. I thought it'd be better to use 'access' as the typ so it means something.

doomspork:
Sounds good @hassox, thanks! 👍

doomspork:
@hassox do you want to merge this before we release a new version for #169?

dustinfarris:
Is this actually used anywhere?

doomspork:
@dustinfarris this is not yet used

doomspork:
@hassox can I merge this guy in for the new release?

PR #169, opened 2016-06-22
Fix elixir 1.4 warnings

milmazz:
This PR also remove a duplicated test and use `ExUnit.CaptureLog` to capture a deprecation message when `:on_failure` option is used.


doomspork:
@milmazz did you mean Elixir 1.3?

mix.exs
```diff
@@ -43,6 +43,7 @@ defmodule Guardian.Mixfile do
   defp deps do
     [{:jose, "~> 1.6"},
      {:poison, ">= 1.3.0"},
+     {:phoenix, "~> 1.2.0-rc.1"},
```
doomspork: Why did you include Phoenix?  I don't think we want to support release candidates, we'll end up having to release new versions frequently to keep up.

milmazz:
@doomspork No, I'm referring to Elixir `1.4.0-dev`, which right now points to the `master` branch.

doomspork:
I don't think we need to worry about supporting 1.4-dev just yet.  I'm 👎 on this for now.  Dev and Release Candidate just make for more releases.

@hassox any thoughts?

milmazz:
@doomspork At least for me is annoying to see all those warnings when I use `Guardian` as a dependency. Also, at some point, we'll need to take care of these issues. Why not take care of this sooner than later?

mix.exs
milmazz: @doomspork Right now I don't have access to the machine where I did the changes, but, if I recall well `mix compile` was failing because there is a dependency in `guardian/phoenix/socket.ex` that requires `Phoenix.Socket.assign/3`. wdyt?

doomspork:
Because it is __1.4-dev__ which means between now and the release of stable 1.4 they may change many of these things requiring another Guardian release.  Chasing changes like this does a disservice to the community when we're forced to keep releasing versions to keep up with changes in a non-stable build.  

We haven't even incorporated `with` yet because we're still supporting people on 1.1.

Just my $0.02.

milmazz:
@doomspork @hassox With the latest changes I think we fix all warnings. Also, we keep compatibility with older Elixir versions. Please let me know what do you think.

milmazz:
@hassox Any feedback about this?

doomspork:
Looking good to me @milmazz, thank you 👍 

@hassox any objections?

hassox:
Hey. Sorry I've been travelling. No objections :D Happy to be forward compatible :D

slashmili:
Are you going bump the version in hex? 😬 
BTW can you also update `credo`? They also fixed the warnings in `0.4.5`

Thanks!

doomspork:
@slashmili please open a PR for `credo` 👍

doomspork:
Yes, sorry for the delays.

PR #170, opened 2016-06-23
Refactor utils.ex with ":os.system_time" rather than ":os.timestamp"

ylankgz:
Seems like there is a built-in method in erlang to fetch seconds from `:os.timestamp` http://erlang.org/doc/man/os.html


ylankgz:
Tests failing with erlang 17

PR #178, opened 2016-07-23
Fix for a Dialyzer type matching

Voronchuk:
Had to change `Map` type to `map()` because of dialyzer errors like
```
guardian_db.ex:103: The specified type for the 3rd argument of after_encode_and_sign/4 (#{}) is not a supertype of 'Elixir.Map', which is expected type for this argument in the callback of the 'Elixir.Guardian.Hooks' behaviour
```


doomspork:
Thank you @Voronchuk!  @hassox any thoughts?

Issue #180, opened 2016-07-29
Exclude some controller actions from authentication with Guardian.Plug.EnsureAuthenticated.


ospaarmann:
It would be great if one could exclude some controller actions from authentication when using Guardian.Plug.EnsureAuthenticated. For example in a UserController which also handles signups on the create action. I don't want it to be protected. Creating a special SignupController seems like a not so nice idea.

Something like this would be great:

```
defmodule MyApp.UserController do
  use MyApp.Web, :controller
  alias MyApp.User

  # Current way
  plug Guardian.Plug.EnsureAuthenticated, handler: MyApp.ErrorController

  # Something like this would be nice:
  plug Guardian.Plug.EnsureAuthenticated, handler: MyApp.ErrorController, except: [:create, :new]

  plug :scrub_params, "data" when action in [:create, :update]

  def create(conn, %{"data" => %{"attributes" => user_params}}) do
    changeset = User.registration_changeset(%User{}, user_params)

    case Repo.insert(changeset) do
      {:ok, user} ->
        conn
        |> put_status(:created)
        |> render("show.json", user: user)
      {:error, changeset} ->
        conn
        |> put_status(:unprocessable_entity)
        |> render(MyApp.ChangesetView, "error.json", changeset: changeset)
    end
  end

  def index(conn, _params) do
    users = Repo.all(User)
    render conn, users: users
  end

  def show(conn, %{"id" => id}) do
    user = Repo.get(User, id)
    render conn, user: user
  end
end
```

AlejandroHuerta:
This took me a while to figure out myself but you can write it this way:

`plug Guardian.Plug.EnsureAuthenticated, [handler: MyApp.SessionController] when not action in [:create, :new]`

doomspork:
@AlejandroHuerta that's a good point.  That is almost certainly a more Elixir way of doing it.

AlejandroHuerta:
I should also add that what I posted works right now. @doomspork

doomspork:
@AlejandroHuerta right 😀  I've been working in Ruby too much lately and spaced on guards, thank you for saying something!  Guard is the more appropriate way to do this.

hopewise:
Writing in this way:

    plug Guardian.Plug.EnsureAuthenticated, [handler: MyApp.SessionController] when not action in [:create, :new]

I got error:

```
== Compilation error on file web/router.ex ==
** (CompileError) web/router.ex:16: undefined function when/2
[...]

PR #181, opened 2016-07-29
Allow to exclude some controller actions from authentication.

ospaarmann:
Add option to Guardian.Plug.EnsureAuthenticated to allow controller actions to be excluded from authentication.

Refers to this [issue](https://github.com/ueberauth/guardian/issues/180).

From Readme:

It is also possible to pass a list of actions to be excluded from authentication.
This can be useful for signup. Here you could exclude the `:new` and `:create` actions
in the UserController (or whatever handles your signup).

```elixir
defmodule MyApp.MyController do
  use MyApp.Web, :controller

  plug Guardian.Plug.EnsureAuthenticated, handler: MyApp.MyAuthErrorHandler, except: [:new, :create]
end
```

### DIFF
--- a/lib/guardian/plug/ensure_authenticated.ex
+++ b/lib/guardian/plug/ensure_authenticated.ex
@@ -43,7 +43,9 @@ defmodule Guardian.Plug.EnsureAuthenticated do
   def call(conn, opts) do
     key = Map.get(opts, :key, :default)
 
-    phoenix_action = Map.get(conn, :private) |> Map.get(:phoenix_action)
+    phoenix_action = conn
+      |> Map.get(:private)
+      |> Map.get(:phoenix_action)
     allowed_actions = Map.get(opts, :except)
 
     if is_list(allowed_actions) && Enum.member?(allowed_actions, phoenix_action) do
--- a/lib/guardian/plug/ensure_authenticated.ex
+++ b/lib/guardian/plug/ensure_authenticated.ex
@@ -16,6 +16,9 @@ defmodule Guardian.Plug.EnsureAuthenticated do
 
       plug Guardian.Plug.EnsureAuthenticated, handler: SomeModule, aud: "token"
 
+      # It is also possible to exclude actions from authentication
+      plug Guardian.Plug.EnsureAuthenticated, handler: SomeModule, except: [:new, :create]
+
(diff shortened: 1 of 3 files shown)


lib/guardian/plug/ensure_authenticated.ex
```diff
@@ -27,22 +30,32 @@ defmodule Guardian.Plug.EnsureAuthenticated do
...
     key = Map.get(opts, :key, :default)
 
-    case Guardian.Plug.claims(conn, key) do
-      {:ok, claims} -> conn |> check_claims(opts, claims)
-      {:error, reason} -> handle_error(conn, {:error, reason}, opts)
-      _ -> handle_error(conn, {:error, :no_session}, opts)
+    phoenix_action = conn
+      |> Map.get(:private)
+      |> Map.get(:phoenix_action)
+    allowed_actions = Map.get(opts, :except)
```
doomspork: `allowed_actions` feels strange to me here.  Could we rename this to something akin to `exempted_actions`?

doomspork:
@hassox / @scrogson care to take a peek at this?  Looks good to me. 👍

doomspork:
@AlejandroHuerta makes a good point in #180 that this could be accomplished with a guard  which is more of the Elixir way of doing it.

```elixir
plug Guardian.Plug.EnsureAuthenticated, [handler: MyApp.SessionController] when not action in [:create, :new]
```

What do you think @ospaarmann?  @hassox or @scrogson any thoughts?

scrogson:
Adding a guard clause would definitely be the right way to do this in Phoenix.

AlejandroHuerta:
I'd just like to mention that if you all decide to implement the :except it would be reasonable to also allow :only. This would be similar to how phoenix does it.

ospaarmann:
@doomspork I agree with that. It seems to be more flexible and doesn't introduce more complexity into the plug itself. 

It would be nice to add an example to the Readme though, so that less experienced users are made aware of this option. I could create a PR for that, if this helps.

doomspork:
@ospaarmann I am going to close this PR but please feel free to open a PR with README updates 😀

lATAl:
hi @doomspork, I don't see any update on this issue. Do I miss something?

PR #205, opened 2016-09-21
Upgrade phoenix and jose

aaronjensen:
Phoenix so we can use it in projects that use phoenix 1.2 w/o override and jose so we can build it with recent elixir/erlang versions

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -45,7 +45,7 @@ defmodule Guardian.Mixfile do
     [{:jose, "~> 1.6"},
      {:poison, ">= 1.3.0"},
      {:plug, "~> 1.0"},
-     {:phoenix, "~> 1.1.0", optional: true},
+     {:phoenix, "~> 1.2.0", optional: true},
      {:ex_doc, "~> 0.10", only: :docs},
      {:earmark, ">= 0.0.0", only: :docs},
      {:credo, "~> 0.3", only: [:dev, :test]},
--- a/mix.exs
+++ b/mix.exs
@@ -42,7 +42,7 @@ defmodule Guardian.Mixfile do
   end
 
   defp deps do
-    [{:jose, "~> 1.6"},
+    [{:jose, "~> 1.8"},
      {:poison, ">= 1.3.0"},
      {:plug, "~> 1.0"},
      {:phoenix, "~> 1.2.0", optional: true},
--- a/mix.exs
+++ b/mix.exs
@@ -15,7 +15,7 @@ defmodule Guardian.Mixfile do
       name: "Guardian",
       app: :guardian,
       version: @version,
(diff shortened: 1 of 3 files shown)


doomspork:
@aaronjensen doesn't look like Jose was updated in your changes.

doomspork:
We should also consider bumping the Elixir version to 1.2: https://github.com/ueberauth/guardian/blob/master/mix.exs#L18

aaronjensen:
@doomspork I had only updated in the `mix.Iock`. I updated it in the `mix.exs` as well and bumped phoenix.

doomspork:
Looks like we'll need to update the TravisCI configuration for this too.  Sorry for the back-and-forth @aaronjensen, would you be willing to make a change to `.travis.yml`?

This will drop testing for Elixir 1.1 and OTP 17.

```yml
language: elixir
elixir:
  - 1.2.5
  - 1.3.4
otp_release:
  - 18.3
  - 19.0 
env: MIX_ENV=test
sudo: false # faster builds
notifications:
  email: false
script:
  - mix compile --warnings-as-errors
  - mix test
  - mix credo
```

aaronjensen:
@doomspork sure thing, how's that?

aaronjensen:
It looks like Elixir 1.2.5 and OTP 19 don't work well together. Want me to remove the old elixir? I think there's a travis syntax for more specific matrixing but I don't know it off the top of my head.

doomspork:
@aaronjensen after playing with it a bit tonight, I believe this is what we need:

```yml
language: elixir
matrix:
  include:
    - otp_release: 18.3
      elixir: 1.3.2
    - otp_release: 19.0
      elixir: 1.3.2
sudo: false
script:
  - mix test
```

aaronjensen:
I added elixir 1.2 in there as well, let's see how it goes.

mix.exs
```diff
@@ -15,7 +15,7 @@ defmodule Guardian.Mixfile do
       name: "Guardian",
       app: :guardian,
       version: @version,
-      elixir: "~> 1.1",
+      elixir: "~> 1.2",
```
doomspork: After conferring with @hassox and @scrogson, we're going to bump this to 1.3 and drop 1.2

doomspork:
@aaronjensen I think we can safely drop 1.2 in favor of 1.3 now that 1.4 is preparing to ship.  I've gone ahead and updated the rest of the projects within the Ueberauth organization.

aaronjensen:
Sounds good, updated.

doomspork:
@aaronjensen I pulled your branch, updated the `.travis.yml` and targeted 1.3, and pushed to master.  Thanks again 👍 

I've released a new version as well.

Issue #210, opened 2016-10-04
Cycling secret key causes error with matching clause {:error, :invalid_token}


jeisenberg:
I noticed when we swapped out the secret_key in the config file, any logged in user would get an application error

`** (CaseClauseError) no case clause matching: {:error, :invalid_token}
lib/guardian/plug/verify_session.ex:29: Guardian.Plug.VerifySession.call/2`

This seems like it could be an easy fix by adding a third case clause in verify_session.ex to 

```
  def call(conn, opts) do
    key = Map.get(opts, :key, :default)

    case Guardian.Plug.claims(conn, key) do
      {:ok, _} -> conn
      {:error, :no_session} ->
        jwt = Plug.Conn.get_session(conn, base_key(key))

        if jwt do
          case Guardian.decode_and_verify(jwt, %{}) do
            {:ok, claims} ->
              conn
              |> Guardian.Plug.set_claims({:ok, claims}, key)
              |> Guardian.Plug.set_current_token(jwt, key)
            {:error, reason} ->
              conn
              |> Plug.Conn.delete_session(base_key(key))
              |> Guardian.Plug.set_claims({:error, reason}, key)
          end
        else
          conn
        end
      {:error, :invalid_token} ->
        conn
        |> Plug.Conn.delete_session(base_key(key))
        |> Guardian.Plug.set_claims({:error, :invalid_token}, key)
    end
  end
```

hassox:
We should probably just catch all errors there rather than just an invalid token one. Did you want to get a PR together?

avocade:
👍

PR #213, opened 2016-10-06
catch all errors on verify_session call instead of just :no_session

jeisenberg:
Pull request in regards to: https://github.com/ueberauth/guardian/issues/210


doomspork:
Thank you @jeisenberg 👍

PR #215, opened 2016-10-20
Fixes name collision issue with Ecto.Query.join/3 in Guardian.Channel

### LINKED ISSUE #214: Name collision using Guardian.Channel and Phoenix
I am using phoenix v1.2.1 with ecto v2.0.5. I don't know if this issue existed before, but if I use `Guardin.Channel` in a phoenix channel I get the following error:

```bash
== Compilation error on file web/channels/visitors_channel.ex ==
** (Ecto.Query.CompileError) malformed join `authed_socket` in query expression
    lib/ecto/query.ex:551: Ecto.Query."MACRO-join"/6
    expanding macro: Ecto.Query.join/3
    web/channels/visitors_channel.ex:3: ExBalance.VisitorsChannel.join/3
```

This is because in `web/web.ex` phoenix imports `Ecto.Query` wich also defines `join/3`.
Currently I am solving this by changing the import to `import Ecto.Query, except: [join: 3] in `web/web.ex` inside the channel function.

Maybe the `join/3` call in `Guardian.Channel's` macro should be more explicit.

kbredemeier:
Resolves #214

### DIFF
--- a/lib/guardian/channel.ex
+++ b/lib/guardian/channel.ex
@@ -49,7 +49,7 @@ defmodule Guardian.Channel do
             |> Map.drop(["guardian_token"])
             |> Map.merge(guardian_params)
 
-            join(room, join_params, authed_socket)
+            __MODULE__.join(room, join_params, authed_socket)
           {:error, reason} -> handle_guardian_auth_failure(reason)
         end
       end


doomspork:
Thank you @kbredemeier, I'll cut a release this evening.

PR #220, opened 2016-10-27
Support {:system, VAR_NAME} in config

### LINKED ISSUE #219: Support {:system, VAR_NAME} idiom for secret_key
It's a pretty common standard to support such a value in config for things that should be loaded from the environment. Secrets and seed values to algorithms are a common thing.

If you agree this would be useful, I can work on this.

michalmuskala:
Closes #219

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -7,6 +7,7 @@ config :guardian, Guardian,
       verify_issuer: true,
       secret_key: "woiuerojksldkjoierwoiejrlskjdf",
       serializer: Guardian.TestGuardianSerializer,
+      system_foo: {:system, "FOO"},
       permissions: %{
         default: [:read, :write, :update, :delete],
         other: [:other_read, :other_write, :other_update, :other_delete]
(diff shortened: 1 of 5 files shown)


michalmuskala:
The failure seems to be cause by credo bug, that refuses to accept code such as:
```elixir
foo() |> bar()
```
The update of credo fixes the issue according to my tests.

doomspork:
@hassox any thoughts on this before I merge?

hassox:
The code looks just fine to me. Travis is erroring though.

michalmuskala:
@hassox yes, travis errors, because credo doesn't like that line https://github.com/ueberauth/guardian/pull/220/files#diff-98bc41bc6745bf48898940055a7dfb7aR276
It looks like a bug, in the credo version that guardian currently uses. Upgrading it fixes this.

doomspork:
@michalmuskala could you bump the credo version so the build passes?

michalmuskala:
@doomspork updated

doomspork:
Thank you @michalmuskala!

Issue #221, opened 2016-10-28
Function references and anonymous functions are invalid in a sys.config


coryodaniel:
I ran across an issue with generating an invalid `sys.config` file using a function reference or anonymous function as per the example:

``` elixir
config :guardian, Guardian,
  allowed_algos: ["Ed25519"],
  secret_key: fn ->
    # some cool way to get the key
  end
```

Got the error when working with [distillery](https://github.com/distillery) but found an [issue on exrm](https://github.com/bitwalker/exrm/issues/230) discussion the problem.

I saw another [ticket](https://github.com/ueberauth/guardian/issues/212) referencing invalid documentation, just figured I'd link it as well. 

I'm working on a fix now that would allow you to pass a tuple to secret key and call `apply/3` on it [here](https://github.com/ueberauth/guardian/blob/master/lib/guardian.ex#L299-L303).

Result will be:

``` elixir
config :guardian, Guardian,
  allowed_algos: ["Ed25519"],
  secret_key: {WayTo, :get, ["my", "key"]}
```

coryodaniel:
PR: https://github.com/ueberauth/guardian/pull/222

doomspork:
Thank you for the report @coryodaniel, I'll take a peek at your PR today 👍

coryodaniel:
@doomspork I added another test case for something I missed ^

PR #222, opened 2016-10-28
Replaces function values with tuples for Guardian :secret_key option

coryodaniel:
``` elixir
def MySecretKey do
  def fetch do
    System.get_env("SECRET_KEY_PASSPHRASE") |> JOSE.JWK.from_file(System.get_env("SECRET_KEY_FILE"))
  end  
end

config :guardian, Guardian,
  allowed_algos: ["Ed25519"],
  secret_key: {MySecretKey, :fetch}
```

Resolves issue #221

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -298,8 +298,9 @@ defmodule Guardian do
 
   defp jose_jwk(the_secret = %JOSE.JWK{}), do: the_secret
   defp jose_jwk(the_secret) when is_binary(the_secret), do: JOSE.JWK.from_oct(the_secret)
-  defp jose_jwk(the_secret) when is_function(the_secret, 0), do: the_secret.()
   defp jose_jwk(the_secret) when is_map(the_secret), do: JOSE.JWK.from_map(the_secret)
+  defp jose_jwk({mod, fun}), do: jose_jwk(:erlang.apply(mod, fun))
+  defp jose_jwk({mod, fun, arity}), do: jose_jwk(:erlang.apply(mod, fun, arity))
   defp jose_jwk(nil), do: jose_jwk(config(:secret_key) || false)
 
   defp encode_claims(claims) do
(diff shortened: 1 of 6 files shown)


lib/guardian.ex
```diff
@@ -298,8 +298,9 @@ defmodule Guardian do
 
   defp jose_jwk(the_secret = %JOSE.JWK{}), do: the_secret
   defp jose_jwk(the_secret) when is_binary(the_secret), do: JOSE.JWK.from_oct(the_secret)
-  defp jose_jwk(the_secret) when is_function(the_secret, 0), do: the_secret.()
   defp jose_jwk(the_secret) when is_map(the_secret), do: JOSE.JWK.from_map(the_secret)
+  defp jose_jwk({mod, fun}), do: jose_jwk(:erlang.apply(mod, fun))
+  defp jose_jwk({mod, fun, arity}), do: jose_jwk(:erlang.apply(mod, fun, arity))
```
shantiii: The third parameter of the tuple is the argument list, not the arity.

doomspork: @coryodaniel it's nitpicky but could you fix this and rename the parameter?

coryodaniel: Yep!

doomspork:
Thank you @coryodaniel!

doomspork:
@hassox I'm all for this change, any thoughts or concerns before I merge and cut a new version?  This _will_ be a breaking change.

hassox:
Ohh I like it! This is great. Thanks for contributing. Could you please update the Changelog to include this new addition?

coryodaniel:
@hassox Want me to add that to the changelog or was that a note for @doomspork when a new release is made... I wasn't sure where to add it

coryodaniel:
Heyoooo, I didn't cover a test cast that had an issue. Just fixed it. Sorry!

doomspork:
@coryodaniel he was referring to you about the changelog but I can do that if you'd rather 😀

coryodaniel:
@doomspork should I put it under the bug section or add a new release version?

doomspork:
Thank you @coryodaniel 👍

Philpax:
FYI, I just spent some time wracking my head over why the example in the repo's readme (i.e. the tuple format) wouldn't work. As I'm using the latest available public release (i.e. _not_ master), I discovered I had to use the old anonymous function syntax (i.e. `secret_key: fn -> keyGenerationHere end`).

To this end, I suggest adding a note to the readme about this change and removing it when it's time for a new release; might prevent others from getting stuck in the quagmire that I did 😅

PR #233, opened 2016-12-15
Fix `from_token` example on README.

vic:
The claims map as returned by `Guardian.decode_and_verify(jwt)` has an string key `"sub"` which must be given to `Guardian.serializer.from_token`.

Updated the README to reflect this, as accessing `claims.sub` would not work.


doomspork:
Thank you @vic 👍

PR #234, opened 2016-12-16
test before_encode_and_sign & after_encode_and_sign hooks

ericsullivan:
verify before_encode_and_sign and after_encode_and_sign hooks are called
pattern match after_encode_and_sign on :ok or {:error, error}

In [GuarbianDB](https://github.com/hassox/guardian_db/blob/master/lib/guardian_db.ex#L74), after_encode_and_sign can return { :error, :token_storage_failure }

```
  def after_encode_and_sign(resource, type, claims, jwt) do
    case Token.create!(claims, jwt) do
      { :error, _ } -> { :error, :token_storage_failure }
      _ -> { :ok, { resource, type, claims, jwt } }
    end
  end
```

This change pattern matches so that the error is return from encode_from_hooked

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -7,6 +7,7 @@ config :guardian, Guardian,
       verify_issuer: true,
       secret_key: "woiuerojksldkjoierwoiejrlskjdf",
       serializer: Guardian.TestGuardianSerializer,
+      hooks: Guardian.Hooks.Test,
       system_foo: {:system, "FOO"},
       permissions: %{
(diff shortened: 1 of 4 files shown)


doomspork:
This is fantastic @ericsullivan, thank you!! 👍 

@hassox / @scrogson any thoughts before I merge this?

ericsullivan:
Thank you :)

doomspork:
Thank _you_ @ericsullivan 👍

PR #235, opened 2016-12-16
add multihook support

ericsullivan:
I added multihook support. I was reviewing changelogs and I saw it was once supported, then quickly removed. The reasons for it not being included might still apply... In my case, I'm building some devise-like modules (timeoutable), and I wanted to hook into the after_sign_in callback, but also have GuardianDB.

This just functions as a pipeline. For those callbacks that might return an error, it stops the pipeline at that point, so ordering of the hooks is important.

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -108,7 +108,17 @@ defmodule Guardian do
   defp encode_from_hooked({:error, _reason} = error), do: error
 
   @doc false
-  def hooks_module, do: config(:hooks, Guardian.Hooks.Default)
+  def hooks_module do
+    case config(:hooks, Guardian.Hooks.Default) do
+      hooks when is_list(hooks) -> Guardian.Multihook
+      hook -> hook
+    end
+  end
+
+  @doc false
+  def hooks_modules do
+    config(:hooks, [Guardian.Hooks.Default])
+  end
 
   @doc """
   Revokes the current token.
--- a/lib/guardian/multihook.ex
+++ b/lib/guardian/multihook.ex
@@ -1,0 +1,54 @@
+defmodule Guardian.Multihook do
+  @moduledoc """
+  Multihook implementation of GuardianHooks.
+  """
+  @behaviour Guardian.Hooks
+
+  def before_encode_and_sign(resource, type, claims) do
+    _before_encode_and_sign(Guardian.hooks_modules, {:ok, {resource, type, claims}})
+  end
+
+  defp _before_encode_and_sign(_, {:error, _reason} = results), do: results
+  defp _before_encode_and_sign([], results), do: results
+  defp _before_encode_and_sign([hook_module | hooks_modules], {:ok, {resource, type, claims}}) do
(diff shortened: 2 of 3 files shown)


doomspork:
Thank you @ericsullivan, I'll go through this later today and leave feedback.

@hassox / @scrogson — thoughts?

hassox:
Hey @ericsullivan 

Thanks for taking the time to put this together. We did indeed used to have it but I removed it. I talked to eric and jose about this quite a bit and they convinced me that it was potentially problematic.

Just to prompt the start of a debate, the reasons was that allowing multiple hook modules can lead to complexity and implicitness. Where does the hook come from? Are there dependencies between the hooks? Should they always run? Is it explicit such that you can determine the behaviour that you're getting?

It was felt at the time that to support multi-hooks, the most explicit and straight forward way to do it would be to have the developer create a module that delegates to other hook modules. This would provide explicitness and allow the behaviour to match the requirements exactly. 

I'm not entirely against this idea at all, but the discussion with jose and eric were pretty convincing. I more intend this response to start a bit of a debate so we can all get a better understanding of what a good way forward is. (I'm hoping that it feels that way)

ericsullivan:
That makes a lot of sense. As I was implementing this I noticed areas of uncertainty, where I felt I was making an arbitrary decision. For instance, why not run all hooks and return a list of errors. I think it might be easier to discuss the use cases I was working on that lead me here. I posted two more issues, but my guess is the discussion here will close them.

I'm building a new application and I need some features that exist in devise:

[Timeoutable](https://github.com/ueberauth/guardian/issues/236): I was going to hook into after_sign_in to set a last_accessed_at key in the session and a plug that would clear the claims if it was expired otherwise update the last_access_at key.

[Confirmable/Approvable](https://github.com/ueberauth/guardian/issues/237): Hook into encode_and_sign and return an error, and a plug that'd call the error handler if they're not confirmed/approved.

I'm guessing you're going to suggest I create my own wrapper around Guardian.Plug.sign_in, but I wanted to verify that :) I was hoping guardian (through its hooks) could provide a standard interface to build on in the same way that Devise provides active_for_authentication, inactive_message, and such. With this multihook module, other extensions could include instruction like, "add it to the hooks list" and you'd be done. Without the glue it'll be up to a developer (or another mix project) to wire them together. I do think thats a better approach though. I'll get started on it :)

hassox:
I don't think I'd suggest wrapping sign in. I feel like a hooks module would likely be probably what you're after. 

As the current lib stands I would suggest writing a hook module that did exactly what you're looking for.

```elixir
defmodule MyHooks do
  ...
   
  def after_sign_in(conn) do
    with conn <- Timeoutable.after_sign_in
            conn <- AnotherThing.after_sign_in
    do
       conn
    end
  end
```

I'd also see the hooks module that you've built as a meta-hook with known behaviour. TBH I'm not close enough to what you're trying to do to give anything specific but I am wondering if this is not enough.

doomspork:
Closing this since we won't be adding mutli-hook support to Guardian at this time.

Issue #78, opened 2015-12-31
Handling Errors in LoadResource


joshwlewis:
Hi y'all,

Just came across some unexpected functionality:

``` elixir
defmodule MyApp.TokenSerializer do
  @behaviour Guardian.Serializer
  alias MyApp.Repo
  alias MyApp.User

  ...

  def from_token(id) do
    case user = Repo.get(User, id) do
      %User{} ->
        {:ok, user}
      nil ->
        {:error, "Unknown resource"}
    end
  end
end
```

My naive impression was that that returning an error here would cause Guardian to abort the request in some fashion. However, the error is swallowed up [here](https://github.com/ueberauth/guardian/blob/master/lib/guardian/plug/load_resource.ex#L29).

I'll likely implement my own plug to verify that the `current_resource` is not nil for now, but I thought it might be nice if Guardian provided something similar. Two ideas:
- Plug.LoadResource could call the `unauthenticated` method on the handler when the serializer returns an error
- A new plug like (Plug.EnsureResource?) that calls the `unauthenticated` method on the handler when the serializer returns an error

I'm happy to take a stab at a PR if you like either idea.

hassox:
Hi @joshwlewis. Perhaps we could pass an option into load resource where when it can't find one it calls the handler? Perhaps even the presence of a handler option would tell it to call some action?

fnimick:
+1 this would be very helpful!

hassox:
@joshwlewis just checking in on this. Did you need some help?

joshwlewis:
Sorry, looks like I let this slide. Let me see if I can get something together later tonight.

hassox:
No problem @joshwlewis I was just going through issues and wanted to touch base.

doomspork:
Howdy @joshwlewis, any updates here? 😀

doomspork:
@hassox / @joshwlewis any updates here?  If there's no movement how about we close?

coryodaniel:
I came across this same issue/assumption last night? Any progress? If someone isn't actively on this, I wouldn't mind working on a solution tomorrow evening. LMK :D

doomspork:
@coryodaniel I say go for it!

coryodaniel:
@doomspork  should load resource have its own off air handler like ensure authenticated?

Also regarding ensure authenticated ... is that the best name for the module since it's technically ensuring the token is verified?

coryodaniel:
[...]

PR #238, opened 2016-12-17
plug to ensure load_resource actually worked

ericsullivan:
When I first used Guardian it was strange that current_resource could be nil after calling the LoadResource plug. It still seems to be an [issue](https://github.com/ueberauth/guardian/issues/78). I decided to go for 3 pull requests and add it.

The discussion suggested using a new plug or updating LoadResource. I used a new plug because I didn't want to change LoadResources current behavior, but it could have gone either way, maybe with the plug looking like this:

```
plug Guardian.Plug.LoadResource, call_handler: true #assumes Guardian.Plug.ErrorHandler
plug Guardian.Plug.LoadResource, handler: MyApp.MyAuthErrorHandler
```

I'm not sure about the default return code or message. I also didn't know if it should call one of the existing error handler methods, but I figured leaving it explicit gives users of the plug the most flexibility. The code / specs / documentation was just copied and a few variables changed, so thanks for the great base to work in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ defmodule MySecretKey do
       |> JOSE.JWK.from_binary
     Redix.stop(conn)
     rsa_jwk
-  end  
+  end
 end
 
 config :guardian, Guardian,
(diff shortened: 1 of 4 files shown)


doomspork:
Thank you @ericsullivan, I'll review this today and leave feedback 👍

doomspork:
Thank you again @ericsullivan 👍

Issue #242, opened 2016-12-25
Throwing "no case clause matching" when use with GuardianDB


yuchunc:
Hi, I upgraded the to the latest Guardian package, and ran in to this exception:

```
** (exit) an exception was raised:
    ** (CaseClauseError) no case clause matching: {:ok, {%ElixirTw.User{__meta__: #Ecto.Schema.Metadata<:loaded, "users">, crypted_password: nil, email: "poverwhelming@gmail.com", id: 1, inserted_at: #Ecto.DateTime<2016-10-01 14:38:47>, name: "Mickey Chen", oauth_infos: #Ecto.Association.NotLoaded<association :oauth_infos is not loaded>, posts: #Ecto.Association.NotLoaded<association :posts is not loaded>, updated_at: #Ecto.DateTime<2016-10-01 14:38:47>}, nil, %{"aud" => "User:1", "exp" => 1485240250, "iat" => 1482648250, "iss" => "ElixirTW", "jti" => "ea6a199e-4e52-4472-96f9-e84ef0ad4dc1", "pem" => %{}, "sub" => "User:1", "typ" => "access"}, "eyJhbGciOiJFUzUxMiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJVc2VyOjEiLCJleHAiOjE0ODUyNDAyNTAsImlhdCI6MTQ4MjY0ODI1MCwiaXNzIjoiRWxpeGlyVFciLCJqdGkiOiJlYTZhMTk5ZS00ZTUyLTQ0NzItOTZmOS1lODRlZjBhZDRkYzEiLCJwZW0iOnt9LCJzdWIiOiJVc2VyOjEiLCJ0eXAiOiJhY2Nlc3MifQ.AZUR4vzOGuYFEW3z1vvms4VZjPGR6EC4XBjL0Hvmax3CSiU_uyLe7CIpUvXpybmo59MTZZN0ouOFvuaznPuQhDcAAMuUVRWgMjQmJRmJUzr0uEbvm3H-JBVZs0tp-FbtT8rLfC2ZuxAvob8G5v8GPpIObWlUO8LOiUaBD0aw2oWudyiq"}}
        (guardian) lib/guardian.ex:100: Guardian.encode_from_hooked/1
        (guardian) lib/guardian/plug.ex:117: Guardian.Plug.sign_in/4
        (elixir_tw) web/controllers/session_controller.ex:51: ElixirTw.SessionController.callback/2
[...]

PR #243, opened 2016-12-25
fix pattern miss match when use with GuardianDB

yuchunc:
Module Guardian line 105, changed `:ok` to `{:ok, _}`

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -102,7 +102,7 @@ defmodule Guardian do
       type,
       claims_from_hook, jwt
     ) do
-      :ok -> {:ok, jwt, claims_from_hook}
+      {:ok, _} -> {:ok, jwt, claims_from_hook}
       {:error, reason} -> {:error, reason}
     end
   end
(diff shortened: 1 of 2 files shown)


yuchunc:
#242

faywong:
This issue has blocked me(freshman to elixir) for two days before i almost give up the guardian db solution.

Anyway, thanks for google & all your efforts. 
In which release will this bugfix will be included?

yuchunc:
I would suggest you rolling back to an earlier version of Guardian. Since I am not the maintainer, I have no ETA on when thhis might get merged.

Also, if you are just trying to run development, you could just go into `/deps` change Guardian's code, then removed the Guardian directory under `/_build`. It will recompile your change, this was how I was sure this was a library issue.

Hope that helps. :)

doomspork:
Thank you @yuchunc, @faywong I'll be releasing a new version this week.  For the time being you can use:

```ruby
{:guardian, github: "ueberauth/guardian"}`
```

collegeimprovements:
I'm getting the same error with the new version also.

doomspork:
Thanks for the report @collegeimprovements 👍

PR #248, opened 2017-01-04
Changed default return value for after_encode_and_sign hook. Fixes #247

### LINKED ISSUE #247: Calling encode_from_hooked causes error after recent commit
Commit 1deca7f8cb11ca8918892e977444599d7ac4f37b changed the pattern matching in guardian.ex line 100 from ```:ok``` to ```{:ok, _}```:
```elixir
case call_after_encode_and_sign_hook(
      resource,
      type,
      claims_from_hook, jwt
) do
      {:ok, _} -> {:ok, jwt, claims_from_hook}
      {:error, reason} -> {:error, reason}
end
```

But the default return value of the ```after_encode_and_sign``` is still just ```:ok```, so we get the error ```no case clause matching: :ok```.
hooks.ex:
```def after_encode_and_sign(_, _, _, _), do: :ok```

NilsLattek:
Fixes #247

### DIFF
--- a/lib/guardian/hooks.ex
+++ b/lib/guardian/hooks.ex
@@ -10,7 +10,7 @@ defmodule Guardian.Hooks do
       def before_encode_and_sign(resource, type, claims) do
         {:ok, {resource, type, claims}}
       end
-      def after_encode_and_sign(_, _, _, _), do: :ok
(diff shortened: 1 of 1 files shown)


doomspork:
Thank you @NilsLattek

PR #254, opened 2017-01-14
Renew session during sign-in

### LINKED ISSUE #186: Should session id be renewed on sign_in/sign_out?
Since guardian uses token stored in the session to authenticate user, should Guardian also renew session id by calling Plug.Conn.configure_session/2 with option renew: true?

I've tried to look into source of Guardian.Plug.sign_in to see if that's happening but I was not able to find it:

``` elixir
  def sign_in(conn, object, type, new_claims) do
    the_key = Map.get(new_claims, :key, :default)
    new_claims = Map.delete(new_claims, :key)

    case Guardian.encode_and_sign(object, type, new_claims) do
      {:ok, jwt, full_claims} ->
        conn
        |> Plug.Conn.put_session(base_key(the_key), jwt)
        |> set_current_resource(object, the_key)
        |> set_claims({:ok, full_claims}, the_key)
        |> set_current_token(jwt, the_key)
        |> Guardian.hooks_module.after_sign_in(the_key)

      {:error, reason} ->
        Plug.Conn.put_session(conn, base_key(the_key), {:error, reason})
    end
  end
```

This seems to be a common advice: https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Renew_the_Session_ID_After_Any_Privilege_Level_Change

doomspork:
This closes #186

### DIFF
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -117,6 +117,7 @@ defmodule Guardian.Plug do
     case Guardian.encode_and_sign(object, type, new_claims) do
       {:ok, jwt, full_claims} ->
         conn
+        |> Plug.Conn.configure_session(renew: true)
         |> Plug.Conn.put_session(base_key(the_key), jwt)
         |> set_current_resource(object, the_key)
         |> set_claims({:ok, full_claims}, the_key)


lib/guardian/plug.ex
```diff
@@ -117,6 +117,7 @@ defmodule Guardian.Plug do
     case Guardian.encode_and_sign(object, type, new_claims) do
       {:ok, jwt, full_claims} ->
         conn
+        |> Plug.Conn.configure_session(renew: true)
```
scrogson: I'm not sure if it makes a difference...but maybe put this after the call to `put_session` ?

doomspork: I was wondering about this myself but I didn't see anything in the docs, I can swap it up.

doomspork: This doesn't appear to matter and I think it reads a bit clearer this way so I'm going to leave it unless @scrogson or @hassox feels strongly about it.

Issue #252, opened 2017-01-13
dialyzer warns for sign_in/4


hykw:
When I use dialyzer, it warns like the following.

```
web/controllers/page_controller.ex:4: Function index/2 has no local return
web/controllers/page_controller.ex:14: The call 'Elixir.Guardian.Plug':sign_in(conn@1::#{},user@1::#{},'token',[]) breaks the contract ('Elixir.Plug.Conn':t(),any(),atom() | 'Elixir.String':t(),#{}) -> 'Elixir.Plug.Conn':t()
```

The functions, I confirmed, it cause the warning are these(both of them are warned).

```elixir
Guardian.Plug.sign_in(conn, user, :token, [])
Guardian.Plug.sign_in(conn, user, :token, perms: %{user: [:normal_login]})
```

I confirmed it in guardian 0.14.2 and dialyxir 0.4.3.
The test code is here: https://github.com/hykw/guardian_dialyzer_specimen

In the following method, sign_in/3 and sign_in/4(the forth param is just %{}) emits no warning, but in case the last one is List or Keyword, it would be warned.

```elixir
defmodule Specimen.PageController do
  use Specimen.Web, :controller

  def index(conn, _params) do
    user = %Specimen.User{}
    perms = %{user: [:normal_login]}

    # OK
    Guardian.Plug.sign_in(conn, user, :token)
    Guardian.Plug.sign_in(conn, user, :token, %{})


    # NG
    Guardian.Plug.sign_in(conn, user, :token, [])
    Guardian.Plug.sign_in(conn, user, :token, perms: perms)

    render conn, "index.html"
  end

end
```

hykw:
The warning I pasted was run in Erlang 18.2.1
In an Erlang 19.2 environment, the warning differed like below

```
web/controllers/page_controller.ex:4: Function index/2 has no local return
[...]

PR #255, opened 2017-01-16
fix after_encode_and_sign dialyzer_error

ericsullivan:
Fixes https://github.com/ueberauth/guardian/issues/252

### DIFF
--- a/lib/guardian/hooks.ex
+++ b/lib/guardian/hooks.ex
@@ -38,7 +38,7 @@ defmodule Guardian.Hooks do
     type :: atom,
     claims :: map(),
     token :: String.t
-  ) :: Plug.Conn.t
+  ) :: {:ok, {term, atom, map, String.t}}
 
   @callback after_sign_in(
     conn :: Plug.Conn.t,


doomspork:
Thank you @ericsullivan!  While we're updating the typespecs we don't we fix `after_sign_in/2`?  Based on the code here it does _not_ return a connection:

https://github.com/ueberauth/guardian/blob/9720eb6fc5278a9945f4235b0488f7480aeaa086/lib/guardian/plug.ex#L179-L184

ericsullivan:
I put additional updates to the typespec in a different PR because dialyzer failed without this change, but wasn't complaining about the other code. Should "mix dialyzer" be added to travis.yml?

I think after_sign_in does return a conn. The default hook implementation returns conn and the api_sign_in also has a method signature that returns conn.

doomspork:
@ericsullivan if you can rebase this to resolve the conflict I'd be happy to merge 👍

ericsullivan:
https://github.com/ueberauth/guardian/pull/256 actually included this fix. Since that was merged first this is no longer needed.

Issue #253, opened 2017-01-14
Dialyzer Callback Return Type - Hooks


asummers:
It seems the hooks module has callbacks listed that only return `{:ok, stuff}` when the code is capable of handling `{:error, reason}` tuples, as well. This is causing Dialyzer to fail for us when we use GuardianDB. I think the solution is simply to add a `| {:error, atom}` to all the callbacks.

asummers:
This is now closed after having #256 merged into master 👍

PR #256, opened 2017-01-16
add typespecs for errors

ericsullivan:
This addresses https://github.com/ueberauth/guardian/issues/253. It includes https://github.com/ueberauth/guardian/pull/255, but I wasn't sure if I got the error types correct. 

I think the error type can be an atom or string. In GuardianDB, it looks like it's always an atom, but the typespec for methods that call the hooks allow both.

```
  @spec decode_and_verify(String.t, map) :: {:ok, map} |
                                            {:error, atom | String.t}

  @spec encode_and_sign(any, atom | String.t, map) :: {:ok, String.t, map} |
                                                      {:error, atom} |
                                                      {:error, String.t}
```

Guardian.Hooks.Test also returns strings, but I wrote that, so can fix it if it should be atoms... Did you want to keep it being both or only {:error, atom}?

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -41,8 +41,7 @@ defmodule Guardian do
   to obtain a value suitable for storage inside a JWT.
   """
   @spec encode_and_sign(any) :: {:ok, String.t, map} |
-                                {:error, atom} |
-                                {:error, String.t}
+                                {:error, any}
   def encode_and_sign(object), do: encode_and_sign(object, @default_token_type, %{})
 
   @doc """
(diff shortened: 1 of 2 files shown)


zachdaniel:
:+1:

asummers:
Thank you for addressing my issue @ericsullivan!

asummers:
Thinking about this for a bit, even if Guardian only supports `atom`, it should probably allow `any`. If you're trying to use a hook to return some custom error, say an `Error` struct, having this be `any` would allow the most future flexibility.

doomspork:
That's a good point @asummers, @ericsullivan would you mind incorporating that change?

ericsullivan:
I added {:error, any}. Then for the methods that returned the results of calling the hooks I updated them. Then some methods that called them... I'm not confident I got them all, but dialyzer passed. This seems like a larger change than I intended. Would it be best to fix the original issue in this PR, then open another one to support any?

asummers:
I don't see any value in splitting that work, personally. Those changes are all of the "boring" variety, so it's unlikely to break something. I think it looks great as is :)

doomspork:
Thank you @ericsullivan! 👍

PR #257, opened 2017-01-17
Cleaned up nested case using with block.

asummers:
Very cosmetic, but I saw that the project compilation was >= 1.2, so I figured I might as well clean it up.

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -285,18 +285,10 @@ defmodule Guardian do
     {secret, params} = strip_value(params, "secret")
 
     try do
-      case decode_token(jwt, secret) do
-        {:ok, claims} ->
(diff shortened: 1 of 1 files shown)


asummers:
The `{:error, reason}` bubbles up because it doesn't match the with, but I can match it explicitly in a with/do/else since the compilation target is 1.3.

```
with {:ok, claims} <- decode_token(jwt, secret),
     {:ok, verified_claims} <- verify_claims(claims, params),
     {:ok, {claims, _}} <- Guardian.hooks_module.on_verify(verified_claims, jwt) do
  {:ok, claims}
else
  {:error, reason} -> {:error, reason}
end
```

doomspork:
Thank you @asummers 👍

PR #265, opened 2017-02-03
Accept binaries for ttl parameters

Moxide:
Hello,
It would be interesting to accept binaries for ttl parameters :
```
{"5", "days"}
```
Now, only this form is accepted :
```
{5, :days}
```
Many other Elixir projects give this opportunity (Ecto, Phoenix...)
This is useful when you want to be able to change config at runtime : one of the available techniques only works with binary parameters, not atom nor integers...


doomspork:
Thank you @Moxide 👍

PR #275, opened 2017-03-07
Allowing for phoenix rc releases

GBH:
Phoenix as at 1.3-rc right now and I can't install guardian with it. Doing `override: true` on this dependency did nothing, but adjusting version string made it install.

```
Failed to use "phoenix" (version 1.3.0-rc.0) because
  guardian (versions 0.14.0 to 0.14.2) requires ~> 1.2.0
```

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -43,7 +43,7 @@ defmodule Guardian.Mixfile do
 
   defp deps do
     [{:jose, "~> 1.8"},
-     {:phoenix, "~> 1.2.0", optional: true},
+     {:phoenix, "~> 1.2.0 or ~> 1.2-rc", optional: true},
      {:plug, "~> 1.0"},
      {:poison, ">= 1.3.0"},
      {:uuid, ">=1.1.1"},


mix.exs
```diff
@@ -43,7 +43,7 @@ defmodule Guardian.Mixfile do
 
   defp deps do
     [{:jose, "~> 1.8"},
-     {:phoenix, "~> 1.2.0", optional: true},
+     {:phoenix, "~> 1.2.0 or ~> 1.2-rc", optional: true},
```
doomspork: Should this be `1.3-rc` and will this require us updating the package for each release candidate?

scrogson: > Doing override: true on this dependency did nothing, but adjusting version string made it install.

Can you be more specific of your usage of the `override: true` option?

You would need to use `override: true` on your applications dependency on `phoenix`.

We may be able to loosen our dependency on `phoenix` to `~> 1.2` which should include anything in between `1.2` to `> 2.0`.

GBH: Oh you know what. `override: true` works when I put on the `phoenix` dependency. Sorry, I'm kinda new to this. But yeah, you'd think that `~> 1.2` would include `1.3-rc`, but it needs to be explicit. I used this as an example to make it work: https://github.com/phoenixframework/phoenix_live_reload/blob/master/mix.exs#L33

I guess this PR can be closed as `override: true` does work when you plug it in the proper place.

scrogson: 👍

PR #280, opened 2017-03-18
Update README.md

josephan:
I think this addition will help beginners get this package up and running more quickly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,16 @@ Alternatively, a `Map`, `Function`, or `%JOSE.JWK{} Struct` may be specified for
 
 See the [key generation docs](https://hexdocs.pm/jose/key-generation.html) from jose for how to generate your own keys.
 
+To get off the ground quickly, simply replace `<guardian secret key>` in your Guardian config with the output of either:
+
+`$ mix phoenix.gen.secret`
+
+or
+
+`iex(1)> JOSE.JWS.generate_key(%{"alg" => "HS512"}) |> JOSE.JWK.to_map |> elem(1) |> Map.take(["k", "kty"])`
+
+After running `$ mix deps.get` because JOSE is one of Guardian's dependencies.
+
 ```elixir
 ## Map ##
 


doomspork:
@josephan I'm not 100% sure this is the easiest method to generate a secret key.  I imagine most Guardian users will be using Phoenix in which case `mix phoenix.gen.secret` is the easiest:

```elixir
$ mix phoenix.gen.secret
Rej/IfETpAhNvueAL/SxQ/G9qnFtwAtSg7rBPFnga6wk7nP6qgkdZFc9vsU3XfhJ
```

josephan:
@doomspork Oh I didn't know about that. Thanks!

casio:
>  I imagine most Guardian users will be using Phoenix in which case mix phoenix.gen.secret

Hm, that at least it wasnt obvious to me : )

So, setting `secret_key` to the string output of `mix phoenix.gen.secret` is sufficient, right?
I was under the impression I'd at least have to put it into some(?) key of some map : )

A little mention of that in the README would indeed help for a quick test / first go with guardian, imho

doomspork:
@casio / @josephan — if either of you thinks the README could be improved feel free to open a PR we can review 😁

josephan:
@doomspork I can reopen this PR then. @casio do you think the addition was helpful?

casio:
I think both are helpful! The `JOSE` generated map, but also the even simpler one for phoenix users.
Something along these lines maybe?

```
To get off the ground quickly, simply replace `<guardian secret key>` above with the output of either:

`$ mix phoenix.gen.secret`

or

`iex(1)> JOSE.JWS.generate_key(%{"alg" => "HS512"}) |> JOSE.JWK.to_map |> elem(1)`
 ```

doomspork:
@josephan if you want to update this to use @casio's suggestion that would be great 👍

josephan:
@doomspork sorry it took a while. Updated and squashed!

doomspork:
No worries @josephan, thanks again for the contribution 😁

PR #282, opened 2017-03-31
Add ability to set claim to look for in LoadResource

### LINKED ISSUE #268: Using a claim other then "sub" in LoadResource
I have a project where I am given a JWT from Auth0. The claim that contains the value used to find a user is not in `sub` so we could not use LoadResource as is. I wanted to know if it makes sense to make the claim that LoadResource looks for customizable? If so, I can make a PR for it.

bryanjos:
fixes #268 

Adds `claim` option to `Guardian.Plug.LoadResource` in order to specify claim to use when looking for value to pass to serializer

### DIFF
--- a/lib/guardian/plug/load_resource.ex
+++ b/lib/guardian/plug/load_resource.ex
@@ -18,6 +18,8 @@ defmodule Guardian.Plug.LoadResource do
     * `:serializer` - The serializer to use to load the current resource from
         the subject claim of the token. Defaults to the result of
         `Guardian.serializer/0`.
+
+    * `:claim` - The claim to look for to pass value to serializer. Defaults to `sub`.
   """
 
   @doc false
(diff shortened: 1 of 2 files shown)


lib/guardian/plug/load_resource.ex
```diff
@@ -48,8 +50,9 @@ defmodule Guardian.Plug.LoadResource do
 
   defp load_resource(claims, opts) do
     serializer = get_serializer(opts)
+    resource_claim = Map.get(opts, :claim, "sub")
 
-    claims |> Map.get("sub") |> serializer.from_token
+    claims |> Map.get(resource_claim) |> serializer.from_token
```
scrogson: Can you move these to multiple lines?

```ex
claims
|> Map.get(resource_claim)
|> serializer.from_token()
```

PR #289, opened 2017-04-23
Move configuration checks to runtime

yogipatel:
Guardian was checking for its configuration and serializer at compile-time, which is a problem when compiling in an environment without a configuration. Permissions were moved to runtime fetching in #88, and I propose we move the configuration check there as well.

As an example of when a situation like this might occur, I was trying to speed up a Docker build by caching the compiled dependencies, using only `mix.exs` and `mix.lock`. Our configuration changes more often than our dependencies, so I do not include the config files for that cache step else any config change will force a recompile of all dependencies. Hope that makes sense.

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -313,7 +313,8 @@ defmodule Guardian do
 
   @doc false
   def config do
-    Application.get_env(:guardian, Guardian)
+    :guardian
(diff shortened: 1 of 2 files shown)


doomspork:
Thank you and welcome @yogipatel!

Issue #273, opened 2017-03-04
Not compatible with Phoenix 1.3


t4t5:
Using the new Phoenix 1.3-rc, I get the following message when I try to install Guardian:

```
Failed to use "phoenix" (version 1.3.0-rc.0) because
  guardian (version 0.14.2) requires ~> 1.2.0
  phoenix_live_reload (version 1.0.8) requires ~> 1.0 or ~> 1.2-rc
  mix.lock specifies 1.3.0-rc.0

** (Mix) Hex dependency resolution failed, relax the version requirements of your dependencies or unlock them (by using mix deps.update or mix deps.unlock). If you are unable to resolve the conflicts you can try overriding with {:dependency, "~> 1.0", override: true}
```

DisruptiveMind:
you can try to just override the phoenix dependency with `override: true` so inside your `mix.exs`:

```elixir
...
defp deps do
    [{:phoenix, "~> 1.3.0-rc", override: true},
end
...
````

doomspork:
Thank you @t4t5.  Since 1.3 is still a ways off, I don't believe we'll be making any changes here just yet. Please try @DisruptiveMind solution.

PR #293, opened 2017-04-30
Allow using Guardian with Phoenix 1.3.0-rc

odarriba:
Just an update to mix deps and updated `mix.lock`.
Also, test executed without issues with Phoenix 1.3.0-rc.


scrogson:
Please see #273 for proper resolution.

PR #297, opened 2017-05-10
Added an encoder to support both JWT and random Tokens

ericsullivan:
I haven't tested this yet, so I highly doubt it works... Before I did that I wanted to ask if this is something that would be useful in Guardian. I figured this PR framework would be good enough to guide that discussion. Basically, JWTs and their merits / drawbacks show up on Hacker News every few weeks. I think there are use cases for both JWTs and random tokens, but Guardian only supports JWT. I was wondering how hard it would be to add a new encoder and it didn't seem too bad. This is one (of many possible implementations) that drops them in a database. An Agent, ETS, Redis, or others could be built too. For drawbacks, this includes extra fields that aren't necessary with a random token, as well as it's not a JWT, which is what Guardian was built to support. It's also worth nothing this is based off GuardianDB, so it's meant to address (some) of the same use cases. Would a smaller library based off the Guardian framework but to deal with this use case be better?


doomspork:
Howdy @ericsullivan!  Thank you for your interest in contributing to Guardian!

In the future please open issues for discussions around new features.  This is not something we’re looking to support at the time and it adds in additional dependencies that are outside the scope of this library.

ericsullivan:
I kind of figured it might be too different, thanks again for taking a look.

PR #298, opened 2017-05-10
support a simple session without JWT

ericsullivan:
Similar to https://github.com/ueberauth/guardian/pull/297, I've been thinking about something like guardian, but without any tokens. For a simple session JWT is pretty heavyweight. I really like the common set of methods, but this might be too far out of Guardians use case. I basically just store the claims in the session and let Plug handle the security. Then there's a plug to load those claims into the private claims. This won't work for all cases (API, Channels, others?), but I think there's a use case for something that provides a simple serializer for storing a serialized resource and permissions in the session.

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -401,7 +401,7 @@ defmodule Guardian do
 
   defp verify_claims(claims, [], _, _), do: {:ok, claims}
 
-  defp build_claims(object, type, claims) do
+  def build_claims(object, type, claims) do
     case Guardian.serializer.for_token(object) do
       {:ok, sub} ->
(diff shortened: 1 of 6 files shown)


doomspork:
Thanks @ericsullivan, please see my comment on #297

ericsullivan:
Ok, thanks for taking a look. I figured Guardian wasn't the place for it but I wanted to check before creating another authentication solution

doomspork:
@ericsullivan there are quite a few authentication libraries already for Elixir, Guardian just happens to be the JWT solution.

https://github.com/h4cc/awesome-elixir#authentication

PR #299, opened 2017-05-10
adjust variable names

ericsullivan:
The method signature was the same in all these cases, but the variable were named differently. None of them were negatively impacting the code.

### DIFF
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -66,8 +66,8 @@ defmodule Guardian.Plug do
   A simple check to see if a request is authenticated
   """
   @spec authenticated?(Plug.Conn.t, atom) :: atom # boolean
-  def authenticated?(conn, type) do
-    case claims(conn, type) do
+  def authenticated?(conn, the_key) do
+    case claims(conn, the_key) do
       {:error, _} -> false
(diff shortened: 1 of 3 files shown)


doomspork:
Hi @ericsullivan, thank you for your continued interest in contributing to Guardian!

We are not looking to change variable names at this time as this may introduce unnecessary conflicts in a large number of changes we are currently undertaking.

If you're interested in contributing I would encourage you to look for existing issues or to create your own so we can discuss the proposed change.

Thanks!

doomspork:
I should add while not as active as we'd like we are available to discuss proposals on Elixir Slack in the #ueberauth channel.

ericsullivan:
Sure, I was just trying to get the variables to align like in the below case:

```
def authenticated?(conn, type) do
   case claims(conn, type) do

...

def claims(conn, the_key \\ :default) do
```

type is actually a key. Similarly the spec said 

```
test "encode_and_sign(object, aud) with ttl" do
```

but the method is

```
def encode_and_sign(object, type), do: encode_and_sign(object, type, %{})
```

It makes sense it won't be helpful if you're refactoring though. Thanks again for reviewing these PRs today.

Issue #309, opened 2017-05-23
V1 - Plugs


hassox:
We need the plugs to come over to the v1 structure. 

There is an opportunity here to do things a little differently to pre-1.0

All plugs should accept options for the `key` position of the credentials but also the current requirement for `handler` seems a bit heavy handed especially with the advent of the FallbackController.

Having a whole bunch of different callbacks to implement also seems a bit weird. Each plug requiring it's own callback seems a bit un-elixir. What would you folks think of a single function on the handler of

`def auth_error(conn, kind, options)`

Where kind is the kind of failure.

hassox:
These are the options I've got so far: 

  ```elixir
  # easiest to create programatically
  plug MyApp.Tokens.Plug, handler: MyApp.AuthErrorHandler, plugs: [
    {Guardian.Plug.VerifySession, [claims: %{"typ" => "access"}]},
    {Guardian.Plug.VerifyHeader, [realm: "Bearer", claims: %{"typ" => "access"}]}
    {Guardian.Plug.EnsureAuthenticated, []}
    {Guardian.Plug.LoadResource, [allow_blank: false]}
  ]
  
  # maybe a pain if phx does something weird with the plug function
  plug MyApp.Tokens.Plug.pipeline handler: MyApp.AuthErrorHandler, do
    plug Guardian.Plug.VerifySession, claims: %{"typ" =>"access"}
    plug Guardian.Plug.VerifyHeader, realm: "Bearer", claims: %{"typ" => "access"}
    plug Guardian.Plug.EnsureAuthenticated
    plug Guardian.Plug.LoadResource, allow_blank: false
  end
 
  # pretty simple but verbose
  plug MyApp.Tokens.Plug, plug: Guardian.Plug.VerifySession,
                          claims: %{"typ" => "access"},
                          handler: MyApp.AuthErrorHandler

  # Very verbose
  plug Guardian.Plug.Executor, impl: MyApp.Tokens,
                               plug: Guardian.Plug.VerifySession,
                               claims: %{"typ" => "access"},
                               handler: MyApp.AuthErrorHandler

  # simple but verbose
  plug Guardian.Plug.VerifySession, impl: MyApp.Tokens,
                                    claims: %{"type" => "access"},
                                    handler: MyApp.AuthErrorHandler

```

hassox:
A possible alternative:

```elixir

### Before anything else
plug Guardian.Plug.put_module(MyApp.Tokens)

[...]

PR #321, opened 2017-05-26
V1 add pipelines

hassox:
closes https://github.com/ueberauth/guardian/issues/309


lib/guardian/plug/pipeline.ex
```diff
@@ -0,0 +1,192 @@
...
+  """
+  defmacro __using__(opts \\ []) do
+    quote do
+      use Plug.Builder
+      import Guardian.Plug.Pipeline
+
+      def init(opts) do
+        otp_app = Keyword.get(opts, :otp_app) ||
+          Keyword.get(unquote(opts), :otp_app)
+        if !otp_app do
```
doomspork: This requires you provide an `:otp_app` but if you configure via the plug that is unnecessary.

PR #322, opened 2017-05-27
Refactor Guardian.Plug and Guardian.Plug.Pipeline

doomspork:


lib/guardian/plug.ex
```diff
@@ -42,182 +42,133 @@ defmodule Guardian.Plug do
...
-        ) :: Plug.Conn.t()
-  def set_current_token(conn, token, options) do
+  @spec authenticated?(Plug.Conn.t, Guardian.opts) :: true | false
+  def authenticated?(conn, opts) do
     key =
       conn
-      |> fetch_key(options)
-      |> token_key()
+      |> fetch_key(opts)
+      |> token_key
```
hassox: Missing () on a lot of these piplined functions.

lib/guardian/plug.ex
```diff
@@ -227,84 +178,65 @@ defmodule Guardian.Plug do
...
-    else
-      Pipeline.current_key(conn) || default_key()
-    end
+  defp sign_out(%{private: private} = conn, impl, :all, opts) do
+    private
+    |> Map.keys
+    |> Enum.map(&key_from_other/1)
+    |> Enum.filter(&(&1 != nil))
+    |> Enum.uniq
+    |> Enum.reduce({:ok, conn}, &clear_key(&1, &1, impl, opts))
```
hassox: I believe this should be `&clear_key(&1, &2, impl, opts)`

lib/guardian/plug/pipeline.ex
```diff
@@ -122,71 +122,69 @@ defmodule Guardian.Plug.Pipeline do
...
+
+      defp raise_error(key),
+        do: raise "Config `#{key}` is missing for #{__MODULE__}"
     end
   end
 
   def call(conn, opts) do
+    [module, error_handler, key] =
+      opts
+      |> Keyword.take([:module, :error_handler, :key])
```
hassox: This is not guaranteed to return the elements in this order :(

lib/guardian/plug/pipeline.ex
```diff
@@ -122,71 +122,69 @@ defmodule Guardian.Plug.Pipeline do
...
+      |> Keyword.values
+
     conn
-    |> maybe_set_item(:guardian_module, Keyword.get(opts, :module))
-    |> maybe_set_item(
-      :guardian_error_handler,
-      Keyword.get(opts, :error_handler)
-    )
-    |> maybe_set_item(:guardian_key, Keyword.get(opts, :key))
+    |> put_module(module)
```
hassox: this changes the behaviour significantly. If those things aren't set they will override the values to nil rather than leaving them alone.

PR #344, opened 2017-06-11
Expand Travis build matrix

samhamilton:
Hi, 

A small update to bump the OTP version and always test the latest patch releases of Elixir

Thanks
Sam


doomspork:
Thanks and welcome @samhamilton!

PR #350, opened 2017-07-03
Fix validate_conditional_tuple spec

odarriba:
The second attribute is missing on the spec, causing an exception while compiling the lib.

PD: Great release! Hoping to see it soon live!

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -727,7 +727,7 @@ defmodule Guardian do
   Provides a way to track an invalid return tuple via a nicer reason
   """
 
-  @spec validate_conditional_tuple({:ok, any} | {:error, any} | any, tuple) :: {:ok, any} | {:error, any}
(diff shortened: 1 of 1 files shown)


lib/guardian.ex
```diff
@@ -727,7 +727,7 @@ defmodule Guardian do
   Provides a way to track an invalid return tuple via a nicer reason
   """
 
-  @spec validate_conditional_tuple({:ok, any} | {:error, any} | any) :: {:ok, any} | {:error, any}
+  @spec validate_conditional_tuple({:ok, any} | {:error, any} | any, tuple) :: {:ok, any} | {:error, any}
```
hassox: Can we update this to be a little more explanatory? `{module, atom}` maybe?

Thanks for catching this!

odarriba: Good one! PR Updated

hassox:
Thankyou!

PR #347, opened 2017-06-26
Update Travis matrix for OTP 20

samhamilton:

PR #351, opened 2017-07-04
Add OTP 20 to Travis matrix

samhamilton:


doomspork:
Thanks @samhamilton!  Do we want to allow failures on OTP 20 via the `allow_failures` option?  As I understand it there's not complete support for it as-of yet with Elixir.

odarriba:
hey!

There is support for it on 1.4.5 and 1.5.0-rc0 (which I don't know if are being tested).
Also, I think it should work pretty well with 1.4 versions for most software 🤔

doomspork:
Great, thanks for the heads up @odarriba!  @hassox do you have any problems with merging this?

scrogson:
I'm confused...isn't this change already in `master`? #347 

Am I blind? I don't see any difference.

samhamilton:
sorry! I didn't spot that.

PR #352, opened 2017-07-05
Add exception in case no secret key is set

odarriba:
While updating our application to use future Guardian 1.0.0, we led to an strange exception of JWT which ended being caused because the secret key was not set (well, it was set but using the old configuration).

This small PR adds a nicer and more clear exception in case no secret key is set.


hassox:
This is great! Thank you.

PR #355, opened 2017-07-12
V1 make plugs plugs

hassox:
@scrogson thoughts after our conversation the other day?

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -712,19 +712,13 @@ defmodule Guardian do
     end
   end
 
-  @doc """
-  Ensures the return result of calling `{mod, func, args}`
-  is a 2 element tuple with `{:ok, any}`  or `{:error, any}`
-  If not, it will return an error with a message detailing
-  the module, function and return that was received
-  """
-
+  @doc false
   def returning_tuple({mod, func, args}) do
     result = apply(mod, func, args)
     case result do
       {:ok, _} -> result
       {:error, _} -> result
-      resp -> {:error, "Invalid return for #{mod}#{func} - #{inspect(resp)}"}
+      resp -> {:error, "Invalid return for #{mod}##{func} - #{inspect(resp)}"}
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/guardian.ex
```diff
@@ -724,16 +713,20 @@ defmodule Guardian do
...
   """
 
-  @spec validate_conditional_tuple({:ok, any} | {:error, any} | any, {module, atom}) :: {:ok, any} | {:error, any}
-  def validate_conditional_tuple({:ok, _} = resp, _),
-    do: resp
-  def validate_conditional_tuple({:error, _} = resp, _),
-    do: resp
-  def validate_conditional_tuple(resp, {m, f}),
-   do: {:error, "Invalid return for #{m}##{f} - #{inspect(resp)}"}
+  def returning_tuple({mod, func, args}) do
```
scrogson: I need to see this used in context of userland code to understand why it's needed to be public. If it's necessary to export then maybe a new name would fit better?

lib/guardian.ex
```diff
@@ -724,16 +713,20 @@ defmodule Guardian do
...
+    result = apply(mod, func, args)
+    case result do
+      {:ok, _} -> result
+      {:error, _} -> result
+      resp -> {:error, "Invalid return for #{mod}#{func} - #{inspect(resp)}"}
```
scrogson: Given `mod` = `Test` and `func` = `test`, this would result in `Testtest`, eh?

Maybe we want `#{mod}.#{func}/#{length(args)}`

lib/guardian.ex
hassox: I'm not intending it to be public outside of guardian, but I do use it in many places inside guardian. @scrogson do you have a suggestion for how to achieve that?

scrogson: If this function is not intended to be public, we should `@doc false`.

hassox:
@scrogson updated

PR #360, opened 2017-07-25
Replace deprecated String.strip/1 with String.trim/1

tmock12:
I have another PR for V1, but I had already done this locally and figured we may as well fix it on master too? 

If we do it on master I guess we can close #359 and then fix V1 on the merge

Strim.trim/1 was implemented in Elixir 1.3 and we currently have `elixir: "~> 1.3",` in the project definition so I think its fine to switch to trim.


doomspork:
Looks good @tmock12, thanks 👍

PR #361, opened 2017-07-26
Use String.trim to fix deprecation warnings

jayjun:


doomspork:
Thank you @jayjun but this was addressed by #360

PR #362, opened 2017-07-26
Incorporate feedback into V1

doomspork:
@hassox as discussed in Slack, here is the feedback I suggested.  I tried to rebase to pull into the latest master changes but there's a number of conflicts in your earlier commits, should we clean those up?


hassox:
Not sure... The V1 branch is a full replacement of master so I don't think we want to rebase it
 Maybe we could cherry pick the parts that should come over.

Changes LGTM!

doomspork:
That sounds good @hassox 👍

PR #365, opened 2017-07-30
Additional refactors

doomspork:


lib/guardian/token/jwt.ex
```diff
@@ -114,9 +114,8 @@ defmodule Guardian.Token.Jwt do
 
   @behaviour Guardian.Token
 
-  alias Guardian.{Config}
+  alias Guardian.{Config, Token.Jwt.Verify}
```
doomspork: Reduce the number of aliases and unnecessary `{ }` around singulars.

lib/guardian/token/jwt.ex
```diff
@@ -311,8 +310,10 @@ defmodule Guardian.Token.Jwt do
   end
 
   defp fetch_secret(mod, opts) do
-    (opts |> Keyword.get(:secret) |> Config.resolve_value() || apply(mod, :config, [:secret_key]))
-    |> case do
+    secret = Keyword.get(opts, :secret)
```
doomspork: I found this pipe to be very difficult to follow especially since it pipes into a `case`.  I think breaking this out makes the code easier to grok.

lib/guardian/token/jwt.ex
```diff
@@ -359,29 +360,23 @@ defmodule Guardian.Token.Jwt do
...
-  defp set_ttl(claims, requested_ttl) do
-    claims |> set_iat() |> set_ttl(requested_ttl)
-  end
+  defp set_ttl(claims, requested_ttl),
+    do: claims |> set_iat() |> set_ttl(requested_ttl)
 
-  defp assign_exp_from_ttl(the_claims, {iat_v, {seconds, unit}}) when unit in [:second, :seconds] do
-    Map.put(the_claims, "exp", iat_v + seconds)
-  end
+  defp assign_exp_from_ttl(the_claims, {iat_v, {seconds, unit}}) when unit in [:second, :seconds],
```
doomspork: Made these inline functions to be consistent with the rest of the file's single line function definitions.

lib/guardian/token/jwt/verify.ex
```diff
@@ -3,47 +3,38 @@ defmodule Guardian.Token.Jwt.Verify do
...
 
-    if verify_issuer do
-      if issuer == iss do
-        {:ok, claims}
-      else
-        {:error, :invalid_issuer}
-      end
-    else
-      {:ok, claims}
+    cond do
```
doomspork: Refactored out the nested if, I feel the `cond` reads better.

lib/guardian/token/jwt/verify.ex
```diff
@@ -3,47 +3,38 @@ defmodule Guardian.Token.Jwt.Verify do
...
+      verify_issuer && issuer == iss -> {:ok, claims}
+      verify_issuer -> {:error, :invalid_issuer}
+      true -> {:ok, claims}
     end
   end
 
   @doc false
+  def verify_claim(_mod, "nbf", %{"nbf" => nil} = claims, _opts), do: {:ok, claims}
```
doomspork: Use pattern matching to handle the `nil` values and avoid an `if`.

PR #366, opened 2017-07-30
Resource encode/decode requires a tuple

doomspork:
I was following the README and ran into an error, turns out these functions need to return `{:ok, result}`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,11 +52,11 @@ defmodule MyApp.Guardian do
   use Guardian, otp_app: :my_app
 
   def subject_for_token(resource, _claims) do
-    to_string(resource.id)
+    {:ok, to_string(resource.id)}
   end
 
   def resource_from_claims(claims) do
-    find_me_a_resource(claims["sub"])
+    {:ok, find_me_a_resource(claims["sub"])}
   end
 end
 ```
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -295,6 +295,10 @@ defmodule Guardian do
 
   alias Guardian.Token.Verify
 
+  defmodule MalformedReturnValueError do
+    defexception [:message]
+  end
+
   defmacro __using__(opts \\ []) do
     alias Guardian.Config, as: GConfig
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -52,11 +52,11 @@ defmodule MyApp.Guardian do
   use Guardian, otp_app: :my_app
 
   def subject_for_token(resource, _claims) do
-    to_string(resource.id)
+    {:ok, to_string(resource.id)}
```
doomspork: @hassox / @scrogson — what about allowing folks to return result directly?  It seems that a common simple solution would be:

```elixir
def resource_from_claims(%{"sub" => sub}), do: Repo.get(User, String.to_integer(sub))

def subject_for_token(resource, _claims), do: to_string(resource.id)
```

But as it is now, this is what we need:

```elixir
def resource_from_claims(%{"sub" => sub}) do
  case Repo.get(User, String.to_integer(sub)) do
    nil -> {:error, "not found"}
    resource -> {:ok, resource}
  end
end

def subject_for_token(resource, _claims), do: {:ok, to_string(resource.id)}
```

lib/guardian.ex
```diff
@@ -718,7 +718,7 @@ defmodule Guardian do
     case result do
       {:ok, _} -> result
       {:error, _} -> result
-      resp -> {:error, "Invalid return for #{mod}##{func} - #{inspect(resp)}"}
+      resp -> raise "Expected `{:ok, result}` or `{:error, reason}` from #{mod}##{func}, got #{inspect(resp)}"
```
doomspork: Incorporated @michalmuskala's feedback and tweaked the message to inform the user what the function expects.  Thoughts?

README.md
hassox: I would feel a bit weird to be honest. Requiring an ok/error tuple provides a more robust API both internally and externally by being able to rely on the shape of the return. It also provides for the developer to determine what an error is/is not. By making the return type ambiguous we move the requirement to determine the ok/error status of the return to all callsites.

The example above is one of the more simple ones but I've used many that deal with straight up strings as well as calling out to an external service. I don't feel like requiring an ok/error tuple is a burden. I'm feeling like ambiguous returntypes would make dealing with this return type propagate to all callsites rather than keeping it focused.

lib/guardian.ex
hassox: Should we have a specific exception for this? I always appreciate specific exceptions when I'm trying to debug.

doomspork: Sure, what would you call the exception?  MalformedReturnValue?

README.md
doomspork: Sounds good 👍

lib/guardian.ex
hassox: The malformed one jumps out :)

doomspork:
Updated @hassox 👍

hassox:
Thanks mate! LGTM

PR #368, opened 2017-08-06
Typespec keyword new_claims for (api_)sign_in

alxgnon:
Encountered a warning from Dialyzer when calling Guardian.Plug.sign_in with keyword arguments. I decided it wouldn't hurt to specify that sign_in and its close relative api_sign_in can receive a keyword list instead of a map.


hassox:
Thanks @alxgnon

PR #373, opened 2017-08-19
Added callback to Guardian.Plug.ErrorHandler behaviour.

### LINKED ISSUE #372: Should already_authenticated/2 be in behaviour callback list for Guardian.Plug.ErrorHandler?
I was going through and `@impl`ing everything in our codebase at work and noticed that when I put `@impl Guardian.Plug.ErrorHandler` above `already_authenticated/2`, I received a compiler warning that it was not part of the behavior. It seems like nothing in the code is too defensive about calling it, so it seems like it might simply be oversight? I will follow up with a PR that adds it, but was just curious if there was some reason it's not there.

asummers:
Resolves #372

### DIFF
--- a/lib/guardian/plug/error_handler.ex
+++ b/lib/guardian/plug/error_handler.ex
@@ -3,6 +3,7 @@ defmodule Guardian.Plug.ErrorHandler do
   A default error handler that can be used for failed authentication
   """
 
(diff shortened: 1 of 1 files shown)


doomspork:
Great, thank you @asummers!

PR #377, opened 2017-08-28
Add an upgrade guide

hassox:
First cut of an upgrade guide. Realise it's rough to start but hoping to get some feedback to make it slick.

### DIFF
--- a/upgrade_guides/0.14.to.1.0.md
+++ b/upgrade_guides/0.14.to.1.0.md
@@ -1,0 +1,345 @@
+# Upgrade guide 0.14.x to 1.0
+
+The move from 0.14.x to 1.0 is significant behind the scenes. Although we've tried to keep backwards compatibility where we could, in order to move Guardian forward we've had to make some breaking changes.
+Guardian is no longer constrained to using only JWT. Although it provides all the JWT behaviour out of the box that it used to, you can now add your own types of tokens.
+
+Dependencies on Phoenix and Plug have been made optional so that Guardian can be used in a more stand-alone manner.
+
+## Implementation
+
+In 0.14.x Guardian was a single, globally unique implementation. In 1.0 you'll need to define a module where all your authentication related items live. You'll also interact with this module rather than directly with Guardian (unless you're a library developer).
+
+The **Implementation** module is the serializer and provides a place for you to put `Hooks`. To create the implementation:
+
+```elixir
+defmodule MyApp.Guardian do
+  use Guardian, otp_app: :my_app
+
+  # ...
+end
+```
+
+## Configuration
+
+Configuration has not changed a lot, but the way that you can define config values has become a lot more useful.
+
+Configuration values can be given using the following types:
+
+* `{MyModule, :func, [:some, :args]}` Calls the function on the module with args
+* any other value
+
+The old values (assuming you're using JWT) are still present, but there are a few new options.
+
+* `token_verify_module` - The module to verify claims. Default `Guardian.Token.Jwt.Verify`
+* `token_ttl` - A map of `%{<token_type> => <ttl_value>}`
+
+Rather than setting configuration for `Guardian` directly, use your implementation module.
+
+0.14.x
+
+```elixir
+config :my_app, Guardian,
+  # options
+```
+
+1.0
+
+```elixir
+config :my_app, MyApp.Guardian,
+  # options
+```
+
+### Serializers
+
+In 0.14.x you defined a serializer. The two functions `from_token`, and `for_token` have been renamed and moved. In 0.14.x you'd create a serializer with these two functions.
+
+The new functions are `subject_for_token` and `resource_from_claims`. These functions are more clearly named and receive more complete arguments.
+
+```elixir
+defmodule MyApp.Guardian do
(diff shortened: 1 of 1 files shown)


upgrade_guides/0.14.to.1.0.md
```diff
@@ -0,0 +1,348 @@
...
+  def resource_from_claims(claims) do
+    {:ok, find_me_a_resource(claims["sub"])}
+  end
+  def resource_from_claims(_claims) do
+    {:error, :reason_for_error}
+  end
+end
+```
+
+The options for the `use` call require an `:opt_app` specification. You can also include your configuration that does not change across environments here if you choose.
```
scrogson: `otp_app`

upgrade_guides/0.14.to.1.0.md
```diff
@@ -0,0 +1,348 @@
...
+```elixir
+Guardian.encode_and_sign(resource, [token_type, claims])
+Guardian.decode_and_verify(token, [claims_to_check])
+```
+
+Becomes:
+
+```elixir
+MyApp.Guardian.encode_and_sign(resource, [claims, options])
+Guardian.decode_and_verify(token, [claims_to_check, options])
```
scrogson: Missing `MyApp` prefix

upgrade_guides/0.14.to.1.0.md
```diff
@@ -0,0 +1,348 @@
...
+
+#### Usage with Plug
+
+The `Guardian.Plug.EnsurePermissions` Plug has been removed. Instead use the `Guardian.Plug.Bitwise` module. This must be used downstream of a Guardian Pipeline
+
+```elixir
+  plug Guardian.Permissions.Bitwise, ensure: %{user_actions: [:books]}
+```
+
+One of is supported (choose one of a list of possible permissions)
```
scrogson: This sentence needs to be reworded.

upgrade_guides/0.14.to.1.0.md
```diff
@@ -0,0 +1,348 @@
...
+end
+```
+
+## Configuration
+
+Configuration has not changed a lot, but the way that you can define config values has become a lot more useful.
+
+Configuration values can be given using the following types:
+
+* `{:system, "FOO"}` - Read from the system environment
```
scrogson: I think we should slim down this number of configuration options.

* `{:system, "FOO"}` seems unnecessary to me now-a-days. People running releases with `distillery` should just use `"${FOO}"` and `REPLACE_OS_VARS=true`

* We should drop the `{module, function}` version and just support `{module, function, args}`. We can support zero arity functions with: `{module, function, []}`

* Anonymous functions should be dropped altogether as they don't work with releases.

doomspork: 100% this 👍

hassox: * distilleries env injection is unavailable on heroku I believe. Forcing folks to use distillery seems like a bad idea to me
* Pretty sure anonymous functions are required because the value given at runtime with a call is resolved the same way. Can remove it from the config area though since agree we should not say include anon functions in your config

scrogson: I'm not suggesting that we force any deployment strategy. I'm simply saying that the `{:system, "FOO"}` pattern was created for people using releases (`System.get_env/1` doesn't work in the config files for releases).

For those using Heroku or similar, they can simply use `System.get_env/1` in the config file and it will work correctly (unless I'm missing something).

> Pretty sure anonymous functions are required because the value given at runtime with a call is resolved the same way.

I'm not following here, sorry. Can you expand on what you mean? This section is talking about configuration, right?

hassox:
@scrogson updated

upgrade_guides/0.14.to.1.0.md
bitwalker: The `{:system, "FOO"}` pattern at this point is an anti-pattern. I would use the `init/2` strategy Phoenix has taken, where the Phoenix endpoint module will have it's `init/2` callback invoked before it starts allowing it to do configuration of the application. If that's not viable, having users provide a tuple of `{module, atom, [args]}` in their configuration which will be invoked as a callback when the application starts to do dynamic config is another option. Anonymous functions in the config is another anti-pattern, so use MFA tuples instead.

doomspork: Thanks @bitwalker 👍

scrogson: @bitwalker I thought about `init/2` as well...it makes sense for a process...but Guardian is all just functions.

bitwalker: @scrogson Yeah, true, though if you need configuration in order to take some action in a function, invoking a function to get that configuration each time may actually be a nice feature, since it naturally supports runtime reconfiguration of the system (and testing too for that matter). Doing so as a callback may not make sense, but supplying an MFA to invoke each time seems like it would be a sane option. That said, I'm not familiar enough with ueberauth to make any specific recommendations, just wanted to throw in my two cents on what I've seen work elsewhere, and what works well with releases.

Issue #375, opened 2017-08-25
Change in the README.md


sebastiandelaroche:
mix phoenix.gen.secret it's deprecated change by phx.gen.secret

doomspork:
@sebastiandelaroche a PR is welcomed to make the change 👍

PR #378, opened 2017-08-28
Update README to new Phoenix command

david-wilson:
Quick doc fix to address #375. Including both versions of command for now since 1.3 is pretty new.


doomspork:
Welcome and thank you @david-wilson 👍

PR #390, opened 2017-09-14
fix plug deps compilation

yordis:


hassox:
This file should not exist in Guardian 1.0. The callbacks are for 0.14.x and are not appropriate anymore. If we want to update it to Guardian 1.0 it should implement only the `auth_errror` callback.

I'm down for either updating it to the `auth_error` callback or just removing it all together.

yordis:
@hassox nevermind then, `guardian_db` is failing because of this but should be fine after you release `1.0`

yordis:
@hassox actually maybe take it for `0.14.x` if you want

PR #393, opened 2017-09-17
Bumps the version of the beta

hassox:
We removed the extraneous error handler file. I'd like to bump this up so we can get an update of guardian db out there.


doomspork:
Go for it @hassox 👍

Issue #401, opened 2017-10-02
Using halt on auth_error


uzarubin:
I would always get `(Plug.Conn.AlreadySentError) the response was already sent` when I tried to use the example `auth_error` as the `error_handler` inside my pipeline. 

It wasn't until I added `|> halt` that the issue was fixed. See below for an example. 

```elixir
defmodule MyApp.AuthErrorHandler do
  import Plug.Conn

  def auth_error(conn, {type, reason}, _opts) do
    body = Poison.encode!(%{message: to_string(type)})
    # send_resp(conn, 401, body)  <--- This line causes the error

    send_resp(conn, 401, body) |> halt # this makes the error go away  
  end
end
```

I was wondering if this is something that I messed up in my own configuration, or there was a forgotten `halt` in the documentation?

oskar1233:
I suggest pasting your configuration as well.

Answering you question: `halt` shouldn't be needed here, docs aren't wrong on this.

uzarubin:
@oskar1233 I was actually following your example on some of these things.  Here is the config from my project:
``` 
defmodule Wire.Guardian.Pipelines.JsonAuth do
  use Guardian.Plug.Pipeline, otp_app: :wire,
                              module: Wire.Guardian,
                              error_handler: Wire.Guardian.ErrorHandler

  plug Guardian.Plug.VerifyHeader, realm: "Bearer"
  plug Guardian.Plug.EnsureAuthenticated
  plug Guardian.Plug.LoadResource, allow_blank: false
end
```

And here is my `router` file:

```
pipeline :authorize_admin do
    plug Wire.Guardian.Pipelines.JsonAuth
  end

 scope "/admin", WireWeb do
    pipe_through :authorize_admin
    post "/invite-tokens/expire", BusinessAdmin.InvitesController, :expire
  end
```

uzarubin:
Going to go ahead and close this.

hauptbenutzer:
I'm sorry to say, this does prevent me from using guardian 1.x with absinthe (or any such API-framework, I would imagine). The problem here is, that in a GraphQL Query some requested fields might require authentication/authorization before resolving, while some don't. To make the distinction, the plug pipeline has to first reach the Absinthe plug. This, due to the `halt` in the Guardian plugs, cannot happen at the moment. 

In fact, there is no way (as far as I can tell) to use the here-mentioned Guardian plugs, and still reach the app logic (absinthe or anything else, really) without manually calling it in the ErrorHandler module.
[...]

PR #403, opened 2017-10-03
Adding halt to the error pipeline

uzarubin:
This PR solves #401 by adding `halt()` to `verify_header.ex` and `verify_session.ex`


doomspork:
@hassox any objections before I merge?

uzarubin:
I don't use session auth, but I noticed that that file also had the same issue? I can fix that if it's actually a problem.

doomspork:
@uzarubin let's fix that as part of this too 👍

uzarubin:
@doomspork done.

hassox:
LGTM Nice catch! :)

PR #404, opened 2017-10-03
Use a nested module in bitwise to provide Plug callbacks conditionally

rbino:
If it is imported in the scope of the module, the import is executed
even if it's inside a false conditional

### DIFF
--- a/lib/guardian/permissions/bitwise.ex
+++ b/lib/guardian/permissions/bitwise.ex
@@ -375,85 +375,92 @@ defmodule Guardian.Permissions.Bitwise do
   end
 
   if Code.ensure_loaded?(Plug) do
-    import Plug.Conn
-    alias Guardian.Plug, as: GPlug
-    alias GPlug.Pipeline
+    defmodule PlugImpl do
+      @moduledoc false
 
-    @doc false
-    @spec init([plug_option]) :: [plug_option]
-    def init(opts) do
-      ensure = Keyword.get(opts, :ensure)
-      one_of = Keyword.get(opts, :one_of)
+      import Plug.Conn
+      alias Guardian.Plug, as: GPlug
+      alias GPlug.Pipeline
 
-      if ensure && one_of do
-        raise ":permissions and a :one_of cannot both be specified for plug #{
-                to_string(__MODULE__)
-              } "
-      end
+      @doc false
+      @spec init([GBits.plug_option]) :: [GBits.plug_option]
+      def init(opts) do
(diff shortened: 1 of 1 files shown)


rbino:
The alternative for the scoped imports would be using the full function name for `halt` (that is the only function that is currently used from `Plug.Conn`) or using an alias and doing something like `PC.halt()`.

Let me know if you prefer this or one of the two alternatives above and I will fix the PR.

rbino:
If I alias `Plug.Conn` the compiler warns me
```
warning: unused alias Conn
  lib/guardian/permissions/bitwise.ex:364

warning: unused alias Pipeline
  lib/guardian/permissions/bitwise.ex:366
```
(the warning is also already there for `GPlug.Pipeline`).

If the warning it's not a problem I will proceed by using an alias.

Another possible solution I came up with could be  moving the Plug related functions in a separate (possibly nested) module, so that they can be entirely wrapped in a conditional (like all the other Plug dependent modules) and do a conditional `defdelegate` from the top module, e.g.
```
  if Code.ensure_loaded?(Plug) do                                                
    defdelegate init(opts), to: Guardian.Permissions.Bitwise.PlugImpl            
    defdelegate call(conn, opts), to: Guardian.Permissions.Bitwise.PlugImpl      
                                                                                 
    defmodule PlugImpl do                                                        
      @moduledoc false                                                           
                                                                                 
      import Plug.Conn                                                           
      alias Guardian.Permissions.Bitwise, as: GBits                              
      alias Guardian.Plug, as: GPlug                                             
      alias GPlug.Pipeline                                                       
                                                                                 
                                                                                 
      @doc false                                                                 
      @spec init([GBits.plug_option]) :: [GBits.plug_option]                     
      def init(opts) do                                                          
        ensure = Keyword.get(opts, :ensure)                                      
        one_of = Keyword.get(opts, :one_of)
       ...         
```

Let me know which of the two approaches you prefer and I will update the PR.

hassox:
To be honest I'd rather not have a separate plug if we can.

rbino:
@hassox would the nested module + `defimpl` be an OK solution (it would not change the current API)?

hassox:
Actually yeah that looks cool.

rbino:
@hassox I've changed the PR to the nested module implementation, let me know if it's ok

doomspork:
This looks good to me @rbino thank you for doing this!  @hassox any final thoughts?  It would be good to get this in.

scrogson:
❤️

PR #409, opened 2017-10-12
Fix typo in Guardian moduledoc

jfornoff:


doomspork:
Thank you @jfornoff 👍

Issue #415, opened 2017-10-31
Fix Guardian compilation when Plug is not being use


yordis:
https://github.com/ueberauth/guardian/blob/708f924806995c749c6c27c3c93dab76fc212bde/lib/guardian/permissions/bitwise.ex#L371  for some reason that line is not actually doing what we expect and the compilation of Guardian fails if there is no `Plug` module

`V1` related

Hanspagh:
So its seems like` Code.ensure_loaded` only work on module level. I propose to move the permission plug to Guardian.Plug.Permission, this however is a breaking change. What do you think about this?

Hanspagh:
Fixed by #404

PR #417, opened 2017-11-07
move plug permisions to plugs

Hanspagh:
Proposed fix to #415. I have moved the permission logic together with the other plugs, since `Code.ensure_loaded ` only seem to work on module level on compile time. I had to move the plug logic into a separate module. This however is a breaking change since plug permissions are no longer available in the module `Guardian.Permissions.Bitwise`


hassox:
This was addressed in https://github.com/ueberauth/guardian/pull/404 but has not yet been released.

I've built against the master branch without plug and without running into this issue.

Can you please confirm that this is still an issue on master?

Hanspagh:
It works, I build from 1.0-beta before but that didn't include #404.

hassox:
Thanks for the PR and confirmation. I will try and get an updated release out tomorrow.

Hanspagh:
Do you track who is working on fixing what issues, should I wouldn't end up trying to fix something that is already fixed :)

hassox:
Unfortunately we don't have a solid way of doing this atm. We know when folks are working on a fix when a PR comes up - but unfortunately this PR didn't close out the bug. 

I'd love to chat about ways we might handle this moving forward in slack #ueberauth. @doomspork will likely have some solid ideas on it also.

Issue #405, opened 2017-10-08
"Remember Me" with a different ttl


mitkins:
I'm trying to utilise the new `remember_me` function. I would like my "refresh" token to last twice as long as my "access" token. In my config I've added a `refresh_ttl` key/value pair:

``` elixir
config :zoinks, Zoinks.Guardian,
  issuer: "Zoinks",
  ttl: { 30, :days },
  refresh_ttl: { 60, :days }
```

In my `SessionController`, I've created the following private function to create the refresh token using the setting (above):

``` Elixir
defp remember_me( conn, user, remember_me ) do
  if ( remember_me ) do
    claims = Zoinks.Guardian.Plug.current_claims(conn)
    refresh_ttl = Zoinks.Guardian.config( :refresh_ttl )

    Zoinks.Guardian.Plug.remember_me( conn, user, claims, [ttl: refresh_ttl, token_type: "refresh"] )
  else
    conn
  end
end
```

Is the above code a reasonable approach?

The `remember_me` function is called right after the sign in:

``` Elixir
conn
|> Zoinks.Guardian.Plug.sign_in(user)
|> remember_me( user, user_params["remember_me"] == "true" )
```

Should I see a third cookie in the browser after the sign in is successful (I currently have 2 - `default` and `_zoinks_key`?

The other thing is that when `Guardian.Plug.VerifyCookie` calls `Guardian.exchange` - will this keep the ttl of the refresh token, instead of using the default ttl?

hassox:
Hey @mitkins For the TTL question, you can specify the TTL of different token types in your config or as options to `use Guardian`.

* `token_ttl` a map of `token_type` to `ttl`. Set specific ttls for specific types of tokens

The option for this would be:

```elixir
  token_ttl: %{
    "refresh" => {14, :days},
    "access" => {7, :days},
  }
```

Or you can set a default with the `ttl` option and just overwrite it.

```elixir
  ttl: {7, :days},
  token_ttl: %{
    "refresh" => {14, :days}
  }
```

For the remember me cookie... Pretty sure that's a bug. They're all getting the same key which is not the intention :(

Thankyou for raising that!

mitkins:
After doing some digging, I found out the following additional information:

* With the default configuration `Guardian.Plug.VerifyCookie` expects to find the refresh token in a cookie called "guardian_default_token". So it doesn't end up finding it in the cookie it was put in ("default")
[...]

Issue #386, opened 2017-09-10
AuthPipeline Phoenix Example


praveenperera:
It might be helpful to have an example of how to use the auth pipeline concept with the phoenix router. Here is what I ended up with, it works but I don't know if its a good way of doing it.

```elixir
defmodule MyApp.Guardian.AuthPipeline.Browser do
  use Guardian.Plug.Pipeline, otp_app: :myapp,
                              module: MyApp.Guardian,
                              error_handler: MyApp.Auth.ErrorHandler

  plug Guardian.Plug.VerifySession
  plug Guardian.Plug.LoadResource, ensure: true, allow_blank: true
end

defmodule MyApp.Guardian.AuthPipeline.JSON do
  use Guardian.Plug.Pipeline, otp_app: :myApp,
                              module: MyApp.Guardian,
                              error_handler: MyApp.Auth.ErrorHandler

  plug Guardian.Plug.VerifyHeader, realm: "Bearer"
  plug Guardian.Plug.LoadResource, allow_blank: true
end

defmodule MyApp.Guardian.AuthPipeline.Authenticate do
  use Guardian.Plug.Pipeline, otp_app: :qchat,
                              module: Qchat.Guardian

  plug Guardian.Plug.EnsureAuthenticated
end
```

and the router.ex file

```elixir
  pipeline :browser_session do
    plug MyApp.Guardian.AuthPipeline.Browser
    plug MyApp.Auth.CurrentUser
  end

  pipeline :login_required do
    plug MyApp.Guardian.AuthPipeline.Authenticate
  end

  pipeline :authorize_admin do
    plug MyApp.Guardian.AuthPipeline.Authenticate
    plug MyApp.Auth.Authorize, :admin
  end

  pipeline :api do
    plug :accepts, ["json"]
    plug MyApp.Guardian.AuthPipeline.JSON
  end
```

doomspork:
@praveenperera we're going to be working on an updated example application for Guardian and Ueberauth to better illustrate the features and implementations 👍 

In the mean time, do you have specific questions?

praveenperera:
@doomspork Sounds good! Nothing specific right now, I got the upgrade done yesterday and it went pretty smoothly. I guess the only thing is, does the AuthPipeline above look okay?

[...]

PR #419, opened 2017-11-07
fix remember_me plug

Hanspagh:
Fix for #405 
Whats need to be done:
- [x] Save the token under the key: guardian_"key"_token, so the verify cookie plug finds it
- [x] Set the secure property on the cookie, so it will only be sent over https
- [x] Set the max_age property on the cookie, should properly be the same as the ttl of the token, is there a convenient way to get the ttl of a token? 
- [x] More tests
- [x] Update docs, see #386  
- [x] Decide if sign_out should also clear the remeber_me cookie
- [x] Decide if we should ignore the token expired error and let Plug.EnsureAuthenticated handle it

### DIFF
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -205,37 +205,37 @@ if Code.ensure_loaded?(Plug) do
     @spec remember_me(Plug.Conn.t(), module, any, Guardian.Token.claims(), Guardian.opts()) ::
             Plug.Conn.t()
     def remember_me(conn, mod, resource, claims \\ %{}, opts \\ []) do
-      with type <- Keyword.get(opts, :token_type, "refresh"),
-           opts <- Keyword.put(opts, :token_type, type),
-           key <- conn |> Pipeline.fetch_key(opts) |> token_key() |> Atom.to_string(),
-           {:ok, token, full_claims} <- Guardian.encode_and_sign(mod, resource, claims, opts) do
-        
-        cookie_opts = cookie_options(mod, full_claims)
-        put_resp_cookie(conn, key, token, cookie_opts)
+      opts = Keyword.put_new(opts, :token_type, "refresh")
+      key = fetch_token_key(conn, opts)
+
+      case Guardian.encode_and_sign(mod, resource, claims, opts) do
+        {:ok, token, new_claims} -> 
+          put_resp_cookie(conn, key, token, cookie_options(mod, new_claims))
+        {:error, _} = err -> 
+          handle_unauthenticated(conn, err, opts)
+      end
+    end
+
+    @spec remember_me_from_token(Plug.Conn.t(), module, Guardian.Token.token(), Guardian.Token.claims(), Guardian.opts()) :: 
+            Plug.Conn.t()
+    def remember_me_from_token(conn, mod, token, claims_to_check \\ %{}, opts \\ []) do
+      token_type = Keyword.get(opts, :token_type, "refresh")
+      key = fetch_token_key(conn, opts)
+
+      with {:ok, claims} <- Guardian.decode_and_verify(mod, token, claims_to_check, opts),
+           {:ok, _old, {new_t, full_new_c}} <- Guardian.exchange(mod, token, claims["typ"], token_type, opts) 
+      do
+        put_resp_cookie(conn, key, new_t, cookie_options(mod, full_new_c))
       else
         {:error, _} = err -> handle_unauthenticated(conn, err, opts)
       end
     end
 
-    @spec remember_me_from_token(
-            Plug.Conn.t(),
-            module,
-            Guardian.Token.token(),
-            Guardian.Token.claims(),
-            Guardian.opts()
-          ) :: Plug.Conn.t()
-    def remember_me_from_token(conn, mod, token, claims_to_check \\ %{}, opts \\ []) do
-      with {:ok, claims} <- Guardian.decode_and_verify(mod, token, claims_to_check, opts),
-           type <- Keyword.get(opts, :token_type, "refresh"),
-           key <- conn |> Pipeline.fetch_key(opts) |> token_key() |> Atom.to_string(),
-           {:ok, _old, {new_t, full_new_c}} <-
-             Guardian.exchange(mod, token, claims["typ"], type, opts) do
-
-        cookie_opts = cookie_options(mod, full_new_c)
-        put_resp_cookie(conn, key, new_t, cookie_opts)
-      else
-        {:error, _} = err -> handle_unauthenticated(conn, err, opts)
-      end
+    defp fetch_token_key(conn, opts) do
+      conn
+      |> Pipeline.fetch_key(opts)
+      |> token_key()
+      |> Atom.to_string()
     end
 
     defp cookie_options(mod, %{"exp" => timestamp}) do
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -42,7 +42,7 @@ if Code.ensure_loaded?(Plug) do
     end
 
     @default_key "default"
-    @default_cookie_options [secure: true, max_age: 60 * 60 * 24 * 7 * 4]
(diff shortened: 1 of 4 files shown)


lib/guardian/plug.ex
```diff
@@ -212,9 +212,9 @@ if Code.ensure_loaded?(Plug) do
     def remember_me(conn, mod, resource, claims, opts) do
       with type <- Keyword.get(opts, :token_type, "refresh"),
            opts <- Keyword.put(opts, :token_type, type),
-           key <- Pipeline.fetch_key(conn, opts),
+           key <- Pipeline.fetch_key(conn, opts) |> token_key() |> Atom.to_string(),
            {:ok, token, _claims} <- Guardian.encode_and_sign(mod, resource, claims, opts) do
-        put_resp_cookie(conn, key, token)
+        put_resp_cookie(conn, key, token, [secure: true, max_age: 1000])
```
hassox: To get the max age here you can extract the expiry from the claims (allow for expiry not being there). It might look something like:

```elixir
@default_cookie_options [secure: true, max_age: 60 * 60 * 24 * 7 * 4]

def remember_me(conn, mod, resource, claims, opts) do
  # snip
  {:ok, token, full_claims} <- Guardian.encode_and_sign(mod, resource, claims, opts)
   
  cookie_opts = cookie_options(mod, full_claims)
  put_resp_cookie(conn, key, token, cookie_opts)
  # snip
end

defp cookie_options(mod, %{"exp" => timestamp}) do
  max_age = timestamp - Guardian.timestamp()
  [max_age: max_age] ++ cookie_options(mod, %{})
end

defp cookie_options(mod, claims) do
  mod.config(:cookie_options, []) ++ @default_cookie_options
end
```

This would give an opportunity to set cookie options in config.

hassox:
I think I'd like to hold this cookie / remember me functionality back fro 1.0 so we can get it out the door then bring it back shortly after once we've had a good chance to kick the tires. What do you think?

Hanspagh:
I am totally fine with delaying it or even closing this PR. Just wanted to notify that I actually started working on this, and get a possible solutions to the max age

Hanspagh:
Okay, I fixed all the things on my to do, expect for 
> Decide if we should ignore the token expired error and let Plug.EnsureAuthenticated handle it #405  

Do you have any opinions on this?

Hanspagh:
I would love to finish this, but I would love a clear decisions on the `token expired error` . Should we just ignore it and assume` Plug.EnsureAuthenticated` will handle it later in the pipeline or does it actually make sense to have `Guardian.Plug.VerifyCookie` throw the error

doomspork:
@Hanspagh I'm open to letting `Plug.EnsureAuthenticated` handle it, I think that keeps everything in one place.  @ueberauth/core any objections to this?

Hanspagh:
I changed the verify cookie plug to not handle the expired token and for now I will skip this `sign_out should also clear the remeber_me cookie`

doomspork:
@Hanspagh is this good to review now?  It appears all the items have been addressed 😀

yordis:
Don't forget to run `mix format` (just remembering you about it 😅)

doomspork:
@yordis I think elixir-lang enforces the formatter has been run by CI?  We should do that if possible 😁

Hanspagh:
@yordis isnt mix format only part of the v1.6.0-dev elixir version?

yordis:
@Hanspagh yes (this is the first programming language that I have no fair for using dev version 😄)

doomspork:
@hassox can you give this another look over before we merge and release?

lib/guardian/plug.ex
```diff
@@ -201,6 +202,51 @@ if Code.ensure_loaded?(Plug) do
       end
     end
 
+    @spec remember_me(Plug.Conn.t(), module, any, Guardian.Token.claims(), Guardian.opts()) ::
+            Plug.Conn.t()
+    def remember_me(conn, mod, resource, claims \\ %{}, opts \\ []) do
+      with type <- Keyword.get(opts, :token_type, "refresh"),
```
scrogson: I'm not sure this is a good use-case for `with`.

This would be better written (IMO) as:

```ex
def remember_me(conn, mod, resource, claims \\ %{}, opts \\ []) do
  opts = Keyword.put_new(opts, :token_type, "refresh")
  key = key(conn, opts)
    
  case Guardian.encode_and_sign(mod, resource, claims, opts) do
    {:ok, token, new_claims} ->
      put_resp_cookie(conn, key, token, cookie_options(mod, new_claims))
    {:error, _} = err ->
      handle_unauthenticated(conn, err, opts)
  end 
end

defp key(conn, opts) do
  conn
  |> Pipeline.fetch_key(opts)
  |> token_key()
  |> Atom.to_string()
end
```

lib/guardian/plug.ex
```diff
@@ -201,6 +202,51 @@ if Code.ensure_loaded?(Plug) do
...
+    end
+
+    @spec remember_me_from_token(
+            Plug.Conn.t(),
+            module,
+            Guardian.Token.token(),
+            Guardian.Token.claims(),
+            Guardian.opts()
+          ) :: Plug.Conn.t()
+    def remember_me_from_token(conn, mod, token, claims_to_check \\ %{}, opts \\ []) do
```
scrogson: Same feedback as above.

Hanspagh:
Requested changes.
English isn't my first language so any feedback on the comments and docs is very welcome.

doomspork:
Thank you @Hanspagh!  @ueberauth/developers any final thoughts before we merge?

robmadole:
@Hanspagh do we need a `remember_me` added to the `__using__` macro so this will get mixed in when defining the Guardian module?

Hanspagh:
@robmadole To be honest, I dont know.

robmadole:
@Hanspagh to enable this from the README:

```
# Set a "refresh" token directly on a cookie.
# Can be used in conjunction with `Guardian.Plug.VerifyCookie`
conn = MyApp.Guardian.Plug.remember_me(conn, resource)
```

I'm 97% sure that it needs to be added to the `Guardian.Plug` `__using__` macro.

robmadole:
Also line 45 where the default cookie stuff is defined:

`@default_cookie_options [secure: true, max_age: 60 * 60 * 24 * 7 * 4]`

Probably should remove the `secure: true` because this will break local developement that isn't running on `https`. `Plug.Conn.put_resp_cookie` will handle setting this correctly according to the documentation.

Hanspagh:
Sure, I can add it.

Hanspagh:
Hi, any status on when is could be accepted?

yordis:
@ueberauth/core  ?

doomspork:
@hassox / @scrogson any final feedback?  You both spent more time reviewing this PR than I did.

robmadole:
We've basically got this running in production for fontawesome.com. So it's had a bit of testing. I'd love to switch over to it once it's released (to get off my local Plugs)

robmadole:
Oh one more thing, if you want to release a pre-release version I'll upgrade our app and give it a test.

doomspork:
That's awesome to hear @robmadole.  I'm merging this so we can cut a release this week 👍

Hanspagh:
Awesome, thank you.

lib/guardian/plug.ex
```diff
@@ -76,6 +77,10 @@ if Code.ensure_loaded?(Plug) do
           do: GPlug.sign_in(conn, implementation(), resource, claims, opts)
 
         def sign_out(conn, opts \\ []), do: GPlug.sign_out(conn, implementation(), opts)
+
+        def remember_me(conn, mod, resource, claims \\ %{}, opts \\ []), do: GPlug.remember_me(conn, mod, resource, claims, opts)
```
gcauchon: How come are `remember_me/5` and `remember_me_from_token/5` not leveraging the `implementation/1` fonction to inject the Guardian implementation module?

Hanspagh: I guess that would make sense. But should properly be in a new PR

PR #421, opened 2017-11-13
Adds a new One time token type.

hassox:
This token type is _not_ a JWT and is useful for when you need one time access.

The one time token uses Ecto to keep track of it's contents if it has been used or not.

If a token has already been used it will be as though that token never existed and will not verify.

From the tests:

```elixir
{:ok, claims} = Impl.decode_and_verify(ctx.token)
assert claims["claims"] == ctx.claims["claims"]

assert Impl.resource_from_claims(claims) == {:ok, %{id: ctx.id}}

assert Impl.decode_and_verify(ctx.token) == {:error, :token_not_found_or_expired}
```

See `Guardian.Token.OneTime` for more information


hassox:
I originally had it in it's own package but put up a q on #ueberauth which came back as put it in core. 

Does it assume postgres? It's my understanding of ecto that it's just going to use whatever your backing DB is, so long as it's an ecto repo. The OneTimeToken requires that you provide a repo to use with it.

scrogson:
@hassox it was my understanding that MySQL didn't support JSON columns? Or does the MySQL adapter handle that automatically as a `text` column and handles casting and dumping automatically?

hassan:
@scrogson MySQL as of 5.7.x supports a JSON datatype.

scrogson:
> @scrogson MySQL as of 5.7.x supports a JSON datatype.

Good to know. Thanks @hassan

scrogson:
Looks like this was merged recently: https://github.com/xerions/mariaex/pull/201

However, it's not yet released on `hex`.

I think we have decided to put this into its own package so that we don't block progress on Guardian 1.0.

PR #422, opened 2017-11-15
Updated README to remove `ensure: true` from Plug example

rosswilson:
The `Guardian.Plug.LoadResource` plug no longer takes an `ensure: true` option, this PR removes that from the README.


hassox:
Thanks for picking this one up!

Can you please change this to `allow_blank` or at least mention it so that folks know it there?

rosswilson:
Thanks @hassox, I've updated the PR with a note to the `allow_blank` option.

doomspork:
Thank you @rosswilson 😀

PR #424, opened 2017-11-18
Remove remember_me functionality

doomspork:


doomspork:
`remember_me` will be added back in via #419 and released as 1.1.0

Issue #414, opened 2017-10-31
use 1.0-beta Got  undefined function find_me_a_resource


zhulinpinyu:
```elixir
defmodule MyApp.Guardian do
  use Guardian, otp_app: :my_app

  def subject_for_token(resource, _claims) do
    {:ok, to_string(resource.id)}
  end

  def subject_for_token(_, _) do
    {:error, :reason_for_error}
  end

  def resource_from_claims(claims) do
    {:ok, find_me_a_resource(claims["sub"])}
  end
  def resource_from_claims(_claims) do
    {:error, :reason_for_error}
  end
end
```

{:guardian, "~> 1.0-beta"}

yordis:
@zhulinpinyu could you put the stack trace? With that people outside of guardian team could take a look faster because we do not know the code base but reading the stack trace we can figure out faster.

zhulinpinyu:
```
➜  demoJWT mix phx.server
Compiling 17 files (.ex)

== Compilation error in file lib/demoJWT/guardian.ex ==
** (CompileError) lib/demoJWT/guardian.ex:13: undefined function find_me_a_resource/1
    (stdlib) lists.erl:1338: :lists.foreach/2
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
    (elixir) lib/kernel/parallel_compiler.ex:121: anonymous fn/4 in Kernel.ParallelCompiler.spawn_compilers/1
```

yordis:
@zhulinpinyu ahhhhhh I see what is the problem

`find_me_a_resource` is not a function from Guardian, the point is that you replaced `find_me_a_resource` with some function from your logic, which could take what comes from `claims["sub"]` and return the resource back. You could do the business logic in that function as well but I wouldn't.

```
 def resource_from_claims(claims) do
    resource = MyApp.get_resource_from_claims(claims["sub"])
    # Do whatever you want but you have to return `{:ok, resource}`
    {:ok,  resource}
  end
```

doomspork:
@yordis is correct, you're referencing a function that does not exist.  Thanks for helping out @yordis ❤️

zhulinpinyu:
@yordis  thanks you much.

```elixir
def resource_from_claims(claims) do
  user = Accounts.get_user!(claims["sub"])
  {:ok, user}
end
```

alec-c4:
Lads, could you add it to documentation?

yordis:
@doomspork I would use this snippet

```elixir
[...]

PR #427, opened 2017-11-24
Update readme example

yordis:
Related to #414

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,15 +52,25 @@ defmodule MyApp.Guardian do
   use Guardian, otp_app: :my_app
 
   def subject_for_token(resource, _claims) do
-    {:ok, to_string(resource.id)}
+    # You can use any value for the subject of your token but
+    # it should be useful in retrieving the resource later, see
+    # how it being used on `resource_from_claims/1` function.
+    # A unique `id` is a good subject, a non-unique email address
+    # is a poor subject.
+    sub = to_string(resource.id)
+    {:ok, sub}
   end
-
   def subject_for_token(_, _) do
     {:error, :reason_for_error}
   end
 
   def resource_from_claims(claims) do
-    {:ok, find_me_a_resource(claims["sub"])}
+    # Here we'll look up our resource from the claims, the subject can be
+    # found in the `"sub"` key. In `above subject_for_token/2` we returned
+    # the resource id so here we'll rely on that to look it up.
+    id = claims["sub"]
+    resource = MyApp.get_resource_by_id(id)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,15 +52,26 @@ defmodule MyApp.Guardian do
   use Guardian, otp_app: :my_app
 
   def subject_for_token(resource, _claims) do
-    {:ok, to_string(resource.id)}
+    # You could send any information you want
```
doomspork: Can we reword this a little differently?  How about:

> You can use any value for the subject of your token but it should be useful in retrieving the resource later.  A unique `id` is a good subject, a non-unique email address is a poor subject.

README.md
```diff
@@ -52,15 +52,26 @@ defmodule MyApp.Guardian do
...
+    {:ok, sub}
   end
-
   def subject_for_token(_, _) do
     {:error, :reason_for_error}
   end
 
   def resource_from_claims(claims) do
-    {:ok, find_me_a_resource(claims["sub"])}
+    # In this case `claims["sub"]` is the `id of the resource`
```
doomspork: How about:

> Here we'll look our resource from the claims, the subject can be found in the `"sub"` key.  In above `subject_for_token/2` we returned the resource `id` so here we'll rely on that to look it up.

README.md
```diff
@@ -52,15 +52,26 @@ defmodule MyApp.Guardian do
...
+    # because we previously used the `id` of the resource for `sub`
+    # of the tokens see `subject_for_token` function.
+    id = claims["sub"]
+    resource = MyApp.get_resource_by_id(id)
+
+    # Do whatever you want for find the resource
```
doomspork: Let's remove this comment.  I think this is applied in the previous comments.

yordis: @doomspork at this point I am commenting for kids (no offenses but people don't like to read and study anymore, they just want to copy + paste)

README.md
```diff
@@ -52,15 +52,25 @@ defmodule MyApp.Guardian do
...
+    {:ok, sub}
   end
-
   def subject_for_token(_, _) do
     {:error, :reason_for_error}
   end
 
   def resource_from_claims(claims) do
-    {:ok, find_me_a_resource(claims["sub"])}
+    # Here we'll look our resource from the claims, the subject can be
```
doomspork: Sorry @yordis, I had a typo in my previous suggestion.  It should be "look up"

doomspork:
Thank you @yordis 👍

Issue #428, opened 2017-11-25
function nil.peek/2 is undefined or private


rafbgarcia:
Just started using Guardian and am facing this issue when attempting to revoke a token.

It says the issue is [in this line](https://github.com/ueberauth/guardian/blob/master/lib/guardian.ex#L664).

(image is clearer, but code is below fwiw)

<img width="1360" alt="2__iex_-s_mix_phx_server" src="https://user-images.githubusercontent.com/1904314/33227195-cd2f7daa-d184-11e7-8e62-aab30e20e0ce.png">

```ex
pry(6)> token
"eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJhcGkiLCJleHAiOjE1MTM5OTc5MzUsImlhdCI6MTUxMTU3ODczNSwiaXNzIjoiYXBpIiwianRpIjoiMmJhNDdiZjEtZTY1Ni00YTFkLTg1ZTctZmI1ODAxYzk3ZmY0IiwibmJmIjoxNTExNTc4NzM0LCJzdWIiOiIyIiwidHlwIjoiYWNjZXNzIn0.78Mr2uBCddb5eBt1tHcXhR2Xebx98sjfh-yvE55l2nAeALtse93CBHC3aUQmJHmPLC_w07vCPaNuoy4D9guqzA"
pry(7)> Api.Guardian.decode_and_verify(token)
{:ok,
 %{"aud" => "api", "exp" => 1513997935, "iat" => 1511578735, "iss" => "api",
   "jti" => "2ba47bf1-e656-4a1d-85e7-fb5801c97ff4", "nbf" => 1511578734,
   "sub" => "2", "typ" => "access"}}
pry(8)> Api.Guardian.revoke(token)
** (UndefinedFunctionError) function nil.peek/2 is undefined or private
[...]

PR #430, opened 2017-11-27
Fix peek function

yordis:
Related to #428


doomspork:
@yordis there is a conflict with the README apparently

yordis:
yeah no clue why I am fixing it

PR #433, opened 2017-11-30
Bad practice to exchange access token for access token

Hanspagh:
I think its bad practice to exchange an access token to an access token. If you want to do that you should properly use the refresh function. Even though it just the upgrade guide, I still think we should teach user the correct way of using the library


doomspork:
@Hanspagh agreed, thank you 👍

Issue #435, opened 2017-12-19
Guardian.revoke(token) method raises error


SofaKing18:
In calling method Guardian.revoke(token) getting error below:

![image](https://user-images.githubusercontent.com/15660128/34160130-be3d9104-e4dc-11e7-9640-502f01d8839e.png)

Seems like `mod.peek(token) ` should be replaced to `token_mod.peek(token)`

But when i did this, got another error:
![image](https://user-images.githubusercontent.com/15660128/34160360-b0919496-e4dd-11e7-81c3-d2dfda565d4e.png)

I'm using Elixir 1.5.2

yordis:
Fixed by https://github.com/ueberauth/guardian/pull/430 

@ueberauth/core could somebody release it if it nos released yet.

yordis:
@hassox Do you think is possible to have some patch release so people dont keep getting into this error?

doomspork:
@yordis and @SofaKing18 v1.0.1 has been released 👍

PR #436, opened 2017-12-19
[#435] Fix keep method call on revoke

SofaKing18:
Hello, I hope I made the right decision, not a Kludge

### DIFF
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -661,7 +661,7 @@ defmodule Guardian do
           {:ok, Guardian.Token.claims()} | {:error, any}
   def revoke(mod, token, opts \\ []) do
     token_mod = Guardian.token_module(mod)
-    with %{claims: claims} <- mod.peek(token),
+    with %{claims: claims} <- token_mod.peek(token_mod, token),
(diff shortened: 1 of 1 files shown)


yordis:
@SofaKing18 this was fixed by https://github.com/ueberauth/guardian/pull/430

SofaKing18:
@yordis 
variable mod is not set => nil.peek(token) raises error
https://github.com/ueberauth/guardian/blob/21622bf30328ef92750dcc5459829016ea4cb93d/lib/guardian.ex#L664

yordis:
@SofaKing18 could you use the `master` branch and verify the error please? This line https://github.com/ueberauth/guardian/blob/2b486689c4605b982951ae56f5a09db1d94b422c/lib/guardian.ex#L393 should fix your issue because you shouldn't be calling directly `.peek` from the module but from Guardian module.

And if the error continues, then could you add unit testing as well so it covers this issue.

SofaKing18:
Error is not actual for master branch, waiting for new release, please :)

yordis:
@ueberauth/core @hassox do you think is possible to get a release for this? Many people that will no setup the module will get this error ...

PR #441, opened 2018-01-03
Improve documentation for verify_header plug module

edouardmenayde:
I'm new to elixir and its ecosystem thus i'm starting to learn its ins and outs including guardian. I love how easy testing is with elixir but I couldn't understand why my test couldn't pass thus I had to dig through the code to understand how the header is parsed by default. I think adding a few line of documentation could help the next guy experience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-Guardian
-========
+# Guardian
 
 > An authentication library for use with Elixir applications.
 
@@ -196,13 +195,13 @@ The default token type of `Guardian` is JWT. It accepts many options but you rea
 
 * `issuer` - The issuer of the token. Your application name/id
 * `secret_key` - The secret key to use for the implementation module.
-                 This may be any resolvable value for `Guardian.Config`
+  This may be any resolvable value for `Guardian.Config`
 
 #### Optional configuration (JWT)
 
 * `token_verify_module` - default `Guardian.Token.Jwt.Verify`. The module that verifies the claims
 * `allowed_algos` - The allowed algos to use for encoding and decoding.
-                    See `JOSE` for available. Default `["HS512"]`
+  See `JOSE` for available. Default `["HS512"]`
 * `ttl` - The default time to live for all tokens. See the type in Guardian.ttl
 * `token_ttl` a map of `token_type` to `ttl`. Set specific ttls for specific types of tokens
 * `allowed_drift` The drift that is allowed when decoding/verifying a token in milliseconds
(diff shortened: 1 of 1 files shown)


oskar1233:
IMHO the stuff you covered is more about understanding  building HTTP request with authorization header than about `verify_header` module itself. Maybe readme or even some separate  guide would be better place to write about it?

edouardmenayde:
@oskar1233 I agree it might be best in the readme. Do I create a new PR or do I amend the existing one ?

doomspork:
@edouardmenayde thanks for getting involved!  I agree with @edouardmenayde that this doesn't really belong in the code documentation, especially not without more context.

Perhaps we should look into adding a section to the README for now on testing best practices?

Thoughts @ueberauth/developers?

oskar1233:
Maybe some subheading in the basics section? Just because that's what you usually do in the beginning and boosts your API understanding :)

edouardmenayde:
Hi!
It's been some days since the last comment thus I think if there's any strong insight against adding that to the readme It would have been raised by now. I took the initiative to amend my last commit and move the documentation to the readme.

README.md
```diff
@@ -387,6 +387,16 @@ This will record each token issued in your database, confirm it is still valid o
 
 For more in-depth documentation please see the [GuardianDb README](https://github.com/ueberauth/guardian_db/blob/master/README.md)
 
+## Best testing practises
```
yordis: Misspelling, change `practises` to `practices`

README.md
```diff
@@ -387,6 +387,16 @@ This will record each token issued in your database, confirm it is still valid o
...
+
+### How to add the token to a request (the Phoenix way)
+```elixir
+{:ok, token, _} = encode_and_sign(resource, %{}, token_type: :access)
```
yordis: I would recommend to add a full example of the Controller so newbies do not get lost in the context.

For example, where `encode_and_sign/2`  come from?
Where should I add this code?

edouardmenayde:
@yordis I fixed the typo and gave some more context to the test. I tried giving much more context with the controller etc. but It become quickly more of a blog post than a little hint to test guardian.

Moggers:
This just saved me from hours of headache. I hadn't used the `Authorization` header before and didn't realize that it had to look like `authorization: <type>: <token>`.

README.md
```diff
@@ -387,6 +392,30 @@ This will record each token issued in your database, confirm it is still valid o
...
+
+### How to add the token to a request (the Phoenix way)
+
+```elixir
+defmodule HelloWeb.AuthControllerTest do
+  use HelloWeb.ConnCase
+  import HelloWeb.Guardian
+
+ test "GET /auth/me", %{conn: conn} do
+    user = insert(:user) // See https://github.com/thoughtbot/ex_machina
```
doomspork: This needs to be `# See https://github.com/thoughtbot/ex_machina`

edouardmenayde: Oh I will correct that :)

edouardmenayde:
@Moggers really happy my headache saved you from having one :)

edouardmenayde:
@doomspork I fixed my mistake, this should be ok to merge :-)

doomspork:
Thank you @edouardmenayde! 👍

PR #443, opened 2018-01-03
Add links to upgrade guide to the documentation

axelson:
It's currently difficult to spot the upgrade guide since it is not linked from either the README or the changelog. This PR makes it easier to find. Here's an example where the user probably didn't notice the upgrade guide: https://elixirforum.com/t/how-to-upgrade-guardian-0-14-to-1-0/11154


axelson:
Cool, thanks!

doomspork:
Thank you @axelson!  If there's any other changes you can think of that would help make the transition better, let us know 👍

PR #444, opened 2018-01-06
Fix incorrect session token key for VerifyCookie

robmadole:
# The fix

This addresses an integration issue between `VerifyCookie` and `VerifySession`.

For these two things to work together they need to agree on the key that is present in `plug_session`.

## VerifySession

Gets the key it uses [here](https://github.com/ueberauth/guardian/blob/master/lib/guardian/plug/verify_session.ex#L73):

```
key = conn |> storage_key(opts) |> token_key()
token = get_session(conn, key)
```

The call to `storage_key` [defers to the Pipeline](https://github.com/ueberauth/guardian/blob/master/lib/guardian/plug/verify_session.ex#L78)

```
Pipeline.fetch_key(conn, opts)
```

And then passes it through [`token_key/1`](https://github.com/ueberauth/guardian/blob/master/lib/guardian/plug/keys.ex#L26)

## VerifyCookie
Puts the session [here](https://github.com/ueberauth/guardian/blob/master/lib/guardian/plug/verify_cookie.ex#L89)

```
put_session(conn, key, token)
```

This `key` comes from [here](https://github.com/ueberauth/guardian/blob/master/lib/guardian/plug/verify_cookie.ex#L91)

```
Pipeline.fetch_key(conn, opts)
```

_What is missing is the call to `token_key/1`_.

Since these two do not agree with one another these Plugs will not work as intended (or at least according to what my understanding of the intention is).

# The test

Currently the test that I wrote is failing when I believe it should be passing. There is an issue with the token that is being generated. I'll keep working on this.

### DIFF
--- a/lib/guardian/plug/verify_cookie.ex
+++ b/lib/guardian/plug/verify_cookie.ex
@@ -63,7 +63,7 @@ if Code.ensure_loaded?(Plug) do
         conn
         |> GPlug.put_current_token(new_t, key: key)
         |> GPlug.put_current_claims(new_c, key: key)
-        |> maybe_put_in_session(active_session?, new_t, key)
+        |> maybe_put_in_session(active_session?, new_t, opts)
       else
         :no_token_found ->
           conn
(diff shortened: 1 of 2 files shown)


doomspork:
Thank you @robmadole!

robmadole:
Kill that test if you like :) And my pleasure.

PR #446, opened 2018-01-15
fix dialyzer error

lukaszsamson:
Fixes
```
lib/my_app_web/auth_pipeline.ex:4: Function raise_error/1 only terminates with explicit exception
```
when compiling auth pipelines with `error_handling` dialyzer option


doomspork:
Thank you @lukaszsamson

PR #454, opened 2018-01-30
Apply mix format

hassox:


config/config.exs
```diff
@@ -15,15 +15,12 @@ use Mix.Config
...
 # by uncommenting the line below and defining dev.exs, test.exs and such.
 # Configuration from the imported file will override the ones defined
 # here (which is why it is important to import them last).
 #
 
-if Mix.env == :test ||
-    Mix.env == :dev ||
-    Mix.env == :docs, do: import_config "#{Mix.env}.exs"
-
+if Mix.env() == :test || Mix.env() == :dev || Mix.env() == :docs,
```
scrogson: Maybe this should be:

```ex
if Mix.env() in [:test, :dev, :docs], do: import_config("#{Mix.env()}.exs")
```

lib/guardian.ex
```diff
@@ -366,9 +366,10 @@ defmodule Guardian do
 
       @spec config() :: Keyword.t()
       def config,
-        do: unquote(otp_app)
-            |> Application.get_env(__MODULE__, [])
-            |> Keyword.merge(unquote(opts))
+        do:
```
scrogson: This should probably just be changed to use the block syntax to avoid this change.

config/config.exs
yordis: @scrogson we do not need to do this.

Nobody will use `prod` on this package and when you use it in another project the `config/config.exs` will be ignore from this package anyway.

Right?!

PR #458, opened 2018-02-07
Call hook to revoke token when using Plug.sign_out

### LINKED ISSUE #447: Guardian.Plug.sign_out doesn't call revoke
Is it intentional that `sign_out` doesn't call revoke on tokens?

tarzan:
The README states that the Plug wrapped around Guardian should revoke tokens when calling `Guardian.Plug.sign_out(conn, opts)`. 

The hook was not actually called though. This PR adds calling the `:on_revoke` hook upon signing out user(s).

fixes #447

### DIFF
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -307,12 +307,9 @@ if Code.ensure_loaded?(Plug) do
     end
 
     defp revoke_token(conn, impl, key, opts) do
-      claims = current_claims(conn, key: key)
       token = current_token(conn, key: key)
 
-      with {:ok, _} <-
-          returning_tuple({impl, :on_revoke, [claims, token, opts]}),
-        do: {:ok, conn}
+      with {:ok, _} <- impl.revoke(token, opts), do: {:ok, conn}
     end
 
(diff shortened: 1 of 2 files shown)


doomspork:
Thank you @tarzan!  We'll look at this and get back to you 👍

doomspork:
@hassox / @scrogson any final thoughts or can we merge this?

lib/guardian/plug.ex
```diff
@@ -298,6 +298,15 @@ if Code.ensure_loaded?(Plug) do
       {:ok, conn}
     end
 
+    defp revoke_token(conn, impl, key, opts) do
+      claims = current_claims(conn, key: key)
+      token = current_token(conn, key: key)
+
+      with {:ok, _} <-
+          returning_tuple({impl, :on_revoke, [claims, token, opts]}),
```
hassox: This should use `impl.revoke(token, opts)`

tarzan:
@hassox I applied the changes you requested :)

hassox:
Thanks @tarzan!

PR #461, opened 2018-02-09
Complement typespec for resource_from_token

pedroassumpcao:
During a project implementation I found out that the typespec for `Guardian.resource_from_token/3` and `Guardian.resource_from_token/4` is incomplete.

The current typespec only accounts for success case but when the token can't be decoded it returns `{:error, any}` from `Guardian.returning_tuple/1`.


scrogson:
:heart:

PR #471, opened 2018-02-22
use implementation macro in remember me

Hanspagh:
Update in reference to the lasted comment on #419


Hanspagh:
Should be good to go now

Issue #469, opened 2018-02-21
Implementing RS256 with Auth0 - multiple secret-keys


mumughal:
Hi,
I am implementing Guardian with Auth0 with RS256 signed JWT tokens. I'm able to make it work. But Auth0 is suggesting that they may be using multiple public keys in the (near) future. Is there a way to support multiple keys in the Guardian module? For example a list of keys? How should this problem be solved?

Thank you so much in advance for your response. :)

yordis:
> using multiple public keys

@mumughal what do you mean, could you elaborate a little bit about what Guardian needs to do for you please.

I dont know if this would help you but https://github.com/ueberauth/guardian/issues/459 check that thread.

mumughal:
Yeah, sure! I get a list of public keys from Auth0 to decode their Rs256 signed JWT. Public keys look something like this:
```{"keys":
[{"alg":"RS256",
"kty":"RSA"
"use":"sig",
"x5c":["___cert___"],
"n":"______",
"e":"AQAB",
"kid":"keyid1",
"x5t":"_______"},
{"alg":"RS256",
"kty":"RSA"
"use":"sig",
"x5c":["___cert___"],
"n":"______",
"e":"AQAB",
"kid":"keyid2",
"x5t":"_______"}]}
```
Now the JWT header has the `kid` for the public signing key of the JWT that should be used for this verifying JWT. For example it looks like this:
```
{
  "typ": "JWT",
  "alg": "RS256",
  "kid": "keyid1"
}
```
Now I have to make sure that guardian module is given the right key, right? Is there any support for this in `Guardian`?
Otherwise the solution I have to consider is that every JWT received should be decoded first just to get the `kid` explicitly, which is not a clean solution :/ What would you recommend?

yordis:
@mumughal I would recommend to follow #459 solution and create your own token for now.

@ueberauth/core any other suggestion here?

mumughal:
[This issue](https://github.com/ueberauth/guardian/issues/459) is referring to a similar problem. But my issue is a bit different. I am talking about decoding a token from a third party service where I have access only to the asymmetric public keys. Please check out the [steps for validating RS256 JWT](https://auth0.com/blog/navigating-rs256-and-jwks/)

mumughal:
@hassox Thank you for addressing the issue. It looks like [this PR](https://github.com/ueberauth/guardian/pull/473) should be able to fix my problem.  :)

hassox:
[...]

PR #473, opened 2018-02-23
JWT - Adds a custom dynamic secret fetcher that can read a tokens header

hassox:
Addresses: https://github.com/ueberauth/guardian/issues/469

### DIFF
--- a/lib/guardian/token/jwt.ex
+++ b/lib/guardian/token/jwt.ex
@@ -168,25 +168,13 @@ defmodule Guardian.Token.Jwt do
 
     defmacro __using__(_opts \\ []) do
       quote do
-        def fetch_signing_secret(mod, opts) do
-          secret = Keyword.get(opts, :secret)
-          secret = Config.resolve_value(secret) || apply(mod, :config, [:secret_key])
+        alias Guardian.Token.Jwt.SecretFetcher.SecretFetcherDefaultImpl, as: DI
 
-          case secret do
-            nil -> {:error, :secret_not_found}
-            val -> {:ok, val}
-          end
-        end
+        def fetch_signing_secret(mod, opts),
+          do: DI.fetch_signing_secret(mod, opts)
 
-        def fetch_verifying_secret(mod, token_headers, opts) do
-          secret = Keyword.get(opts, :secret)
-          secret = Config.resolve_value(secret) || mod.config(:secret_key)
-
-          case secret do
-            nil -> {:error, :secret_not_found}
-            val -> {:ok, val}
-          end
-        end
+        def fetch_verifying_secret(mod, token_headers, opts),
+          do: DI.fetch_verifying_secret(mod, token_headers, opts)
 
         defoverridable fetch_signing_secret: 2, fetch_verifying_secret: 3
       end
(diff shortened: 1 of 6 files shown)


hassox:
@yordis updated

hassox:
@yordis ping

apognu:
Does `fetch_signing_secret` only returns the secret? Signing tokens with different secrets, but not being able to identify which one from the token defeats the purpose, somewhat.

`fetch_signing_secret` should return the secret **and** a key id that would appear in the generated token header as a ```kid``` claim.

Right now, there is no way, from the token headers passed to ```fetch_verifying_secret```, to fetch the key used to sign a token.

scrogson:
@apognu in general, I agree with you. If you are doing public/private key signing/verifying and rotating keys then what you are saying makes perfect sense.

The default JWT token module that ships with Guardian is more naive and doesn't really go to great lengths to support multiple secret keys.

The great news is that you can easily create your own token strategy and use that instead of the default one provided by Guardian.

Let me know if we're missing something.

apognu:
Thanks for replying on a closed issue.

I agree that the default behavior should be a naive one, and while Guardian does allow me to specify custom `secret_fetcher` and `token_verify_module` (if there is another plugging point, maybe I missed it), I believe they lack specific features that would be useful:

 * `fetch_verifying_secret` passes in the JWT headers, so we are able to look for the proper key
 * `fetch_signing_secret` has nothing to do with the JWT headers, so we cannot set, for instance, the `kid`

My specific use-case aside, I believe there is a symmetry issue between those two functions, where one can use the values in the headers and the second cannot set those values.

On the same note (but not directly related to this issue), in custom `token_verify_module`s, the token header is not passed to the `verify_claim` methods, so we still cannot map any verification rules to the key that is used (for example, "a token signed by a specific key can only include the `sub` to which the key belongs to").

hassox:
@apognu I don't believe this has been released yet. I'd be interested to see what modifications you would put in place to make this feature more useful.

scrogson:
@apognu I see. Excellent points! I think it would be great to support these things. Perhaps we can extend what's in master before the next release.

apognu:
@hassox @scrogson I could give it a go, but by the time I familiarize myself with the structure of the project, it could be some time. When is the next release scheduled?

Without having looked too closely at the code, this would require:

 * passing the token headers left and right, so some or all headers are available at different steps of the processing (including in `token_verify_module`;
 * allowing the `secret_fetcher` to modify the header claims (or a subset thereof; should we allow this to dynamically determine the `alg`, on top of the `kid` and secret key?).

Those are my thoughts right now, but I have not spent a lot of time working on it. And I'll be frank, I have no idea how much work that would represent on Guardian's code base.

lib/guardian/token/jwt.ex
```diff
@@ -124,6 +146,66 @@ defmodule Guardian.Token.Jwt do
...
+          do: DI.fetch_verifying_secret(mod, token_headers, opts)
+
+        defoverridable fetch_signing_secret: 2, fetch_verifying_secret: 3
+      end
+    end
+  end
+
+  defmodule SecretFetcher.SecretFetcherDefaultImpl do
+    @moduledoc false
+    use Guardian.Token.Jwt.SecretFetcher
```
ps-feng: I think it's possible to get rid of the macro altogether, but please excuse me if this makes no sense as I've just started learning Elixir and using Guardian to verify Firebase tokens...

Isn't it enough with `@behaviour Guardian.Token.Jwt.SecretFetcher`? After all `fetch_secret_fetcher` is already referencing `SecretFetcherDefaultImpl` by default. This way we can get rid of the macro, which also resembles OO inheritance.

Issue #478, opened 2018-03-03
Custom TTL not working as documented


zapient:
```
# Customize the time to live (ttl) of the token
{:ok, token, claims} = MyApp.Guardian.encode_and_sign(resource, %{}, token_ttl: {1, :minute})
```

Looks like the documentation does not match the [implementation](https://github.com/ueberauth/guardian/blob/master/lib/guardian/token/jwt.ex#L414). Using `ttl` instead of `token_ttl` works as expected.

doomspork:
@zapient a PR to correct the documentation is welcome 👍

zapient:
Sure. Would you like the documentation to match the implementation or the other way around?

yordis:
@zapient fix the documentation.

PR #479, opened 2018-03-04
[Bug #478] Custom ttl incorrectly documented

zapient:
[Bug #478] Fix  custom ttl incorrectly documented


doomspork:
Thank you @zapient 👍

PR #484, opened 2018-03-14
Fix unused variable warning

janpieper:
```
warning: variable "token_headers" is unused
  lib/guardian/token/jwt.ex:198
```


doomspork:
Thank you @janpieper 👍

Issue #483, opened 2018-03-12
see 'Guardian.Plug.Errors'


voger:
In [this line](https://github.com/ueberauth/guardian/blob/7e839ba1c39aed2c303e01f804451cce0f9f924c/lib/guardian/plug/pipeline.ex#L64) the documentation mentions 
> see 'Guardian.Plug.Errors'

but there is not such module. What does it mean?

hlhr:
Just to help out, the option is used to define a ErrorHandler-Module-Pug. There is not predefined plug in `guardian` but you can find an example [here](https://hexdocs.pm/guardian/readme.html#plug-error-handlers) or a definition under `### Error handler` in the same file.

voger:
Thanks. I thought there was indeed somewhere a module with that name and I couldn't find it.

PR #486, opened 2018-03-17
Enhance documentation for error_handler option

hlhr:
As shown in Issue #483 the documentation around the `error_handler` isn't that clear. I had the same issue when I used `Guardian` the first time.

Maybe my changes makes it a little bit clearer, feedback is appreciated.


doomspork:
Thank you @hlhr 😁

Any other documentation enhancements you can think of 👍

PR #491, opened 2018-04-03
Fix upgrade docs load resource allow blank option

alex88:
It seems in 1.0 to ensure the resource found is not nil the option is `allow_blank` right? I cannot find any reference of the `ensure` option


doomspork:
Thank you and welcome @alex88!  Any other improvements to the documentation you can think of would be welcomed 👍

PR #494, opened 2018-04-17
Improve travis setup

yordis:
@ueberauth/developers I am still working on it.


doomspork:
@yordis I'm open to dropping the other versions for Travis.  @hassox thoughts?

hassox:
Always down for dropping old versions personally ;)

yordis:
@ueberauth/core yeah I wouldn't train the folks that it is fine for them to do not do their job upgrading their platform and by so we carry on their responsibilities

yordis:
@ueberauth/core done!

PR #506, opened 2018-06-11
Remove Sentinel from README.md

victor-am:
Since Sentinel has a deprecation notice on it's github page this commit removes it from the Related Projects section of the readme.

Another option would be to simple add a `deprecated` warning on the readme and keep it there. If this is a more desirable option please give me a heads up and I will update the PR 😉 

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,7 +432,6 @@ end
 ## Related projects
 
 * [GuardianDb](https://hex.pm/packages/guardian_db) - Token tracking in the database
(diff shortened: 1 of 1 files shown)


yordis:
@ueberauth/core we can't control that from happening. I would suggest to only leave the Guardian DB package and create a Wiki that we can point to from the readme about related projects so we do not need to commit these changes.

What do you think?

doomspork:
Great idea @yordis 👍

PR #509, opened 2018-06-28
Fix escaped backslash

aditya7iyengar:
Just a minor fix in the documentation


yordis:
@aditya7iyengar thanks lot of the contribution.

I think the fix for this is adding the sigil ` ~S` at the beginning of the module docs, like

```
 @moduledoc ~S"""
....
"""
```

Don't quote me on it, but could you add that change and try it out and see if the issue persist?! (without your changes)

PR #522, opened 2018-08-27
Updating context module name

pera:
In a previous version of this document the "UserManager" context was named "Auth". This commit updates some directories and file names to reflect that change.


pera:
Please let me know if there is any issue with this PR.

PR #524, opened 2018-09-01
Clean up some documentation

ckhrysze:


lib/guardian.ex
```diff
@@ -29,7 +29,7 @@ defmodule Guardian do
   Any other option provided will be merged with the configuration in the config
   files.
 
-  It contains some generated functions, and some callbacks
+  The Guardian module contains some generated functions and some callbacks.
```
yordis: I believe that the comma goes there.

Don't quote me on it because I am not a native English speaker.

But I learned from Grammarly that it is good to add that comma there so you know it is part of the listing rather than grouping.

For example A and B,  C, and D

Replace A, B, C, and D with something that makes sense.

lib/guardian.ex
```diff
@@ -83,7 +83,7 @@ defmodule Guardian do
   * `claims` - Any custom claims that you want to use in your token
   * `opts` - Options for the token module and callbacks
 
-  For more information on options see the documentation for your Token Module.
+  For more information on options see the documentation for your TokenModule.
```
yordis: I don't believe this shouldn't be referring to any naming in your code.

People will try to find `TokenModule` in their code for this change for sure and then open issues because they can't find the module.

lib/guardian.ex
```diff
@@ -115,26 +115,26 @@ defmodule Guardian do
...
   {:ok, claims} = MyApp.Guardian.decode_and_verify(token)
 
   # Decode and verify with literal claims check.
-  # If the cliams in the token do not match those given verification will fail
+  # If the claims in the token do not match those given verification will fail
   {:ok, claims} = MyApp.Guardian.decode_and_verify(token, %{match: "claim"})
 
   # Decode and verify with literal claims check and options.
-  # Options are passed to your Token Module and callbacks
+  # Options are passed to your TokenModule and callbacks
```
yordis: Same here about TokenModule

lib/guardian.ex
```diff
@@ -115,26 +115,26 @@ defmodule Guardian do
...
   {:ok, claims} = MyApp.Guardian.decode_and_verify(token, %{match: "claim"}, some: "secret")
   ```
 
   #### `revoke(token, opts \\ [])`
 
   Revoke a token.
 
-  *Note:* this is entirely dependant on your Token Module and implementation
+  *Note:* this is entirely dependent on your TokenModule and implementation
```
yordis: Same here about the TokenModule

yordis:
@ueberauth/core I would like one of your to give some review as well.

lib/guardian.ex
doomspork: This change is fine 👍

lib/guardian.ex
doomspork: This change isn't necessary.  We're not referencing a specific `TokenModule` rather a `Token Module` that they will create.  Probably doesn't need to be capitalized?

ckhrysze:
I did another pass through, changing the plain text "TokenModule" uses to "token module", but leaving the <TokenModule>.some_function references.

doomspork:
Thank you @ckhrysze 😁

PR #528, opened 2018-10-05
fix guides grouping in docs

Apelsinka223:


doomspork:
We welcome any improvements you can think of!  Documentation could just much love.

Apelsinka223:
You still need to update docs with `mix hex.publish docs`, it's still broken version of guides on hexdocs.pm

PR #537, opened 2018-11-16
Document Revision

evadne:
- fixes a couple of titles and navigation item names


yordis:
@evadne thanks a lot for the contribution 💟

PR #540, opened 2018-11-30
Remove duplicate callback name from docs.

asummers:


yordis:
@asummers thanks a lot 💓

PR #542, opened 2018-12-01
Add error handler behaviour.

asummers:
I found that the documentation around this was kind of scarce, and I had to kind of read the upgrade guide a few times, and guess at the typespec on my own end, so I thought adding a behaviour for documenting the callback would be nice. 

Before I get much further with docs and such, 

1) Is this something you would be into?
2) Is this an appropriate name?
3) Is there anything else you would see fitting into this module that I should move over right away / any existing docs that might be appropriate to grab for moduledocs?

### DIFF
--- a/lib/guardian/error_handler.ex
+++ b/lib/guardian/error_handler.ex
@@ -1,0 +1,4 @@
+defmodule Guardian.ErrorHandler do
+
+  @callback auth_error(conn :: Plug.Conn.t(), {atom, atom}, Guardian.options()) :: Plug.Conn.t()
+end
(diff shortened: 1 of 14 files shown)


doomspork:
Howdy @asummers, thank you for getting involved!  I think this is a good idea we should explore further.  I've added the other contributors so they too can take a peek 👍 

As for documentation:  We _always_ welcome more.

asummers:
I would also say I anticipate there being a few more behaviours that exist that are currently not concrete, so I'm not sure if there should be some folder naming with that thought in mind (e.g. the likely suboptimal subfolder name of `/behaviours/`).

lib/guardian/error_handler.ex
```diff
@@ -0,0 +1,40 @@
+if Code.ensure_loaded?(Plug) do
+  defmodule Guardian.ErrorHandler do
```
yordis: Since this is related to `Plug` I would change the scope to be under `Guardian.Plug` like `Guardian.Plug. ErrorHandler`

PR #543, opened 2018-12-02
Remove Poison from dependencies, replace with Jason in :dev, :test

asummers:


yordis:
@doomspork meh, I would do it since major packages are using Jason by default

asummers:
Re: discussion from offline, the discussion was to add a toggle between Poison and Jason.... But taking a look through the code, this is all just a test dependency, so there's absolutely no reason to add that =) Instead, removing Jason AND Poison from runtime dependencies is the way to go here. 

```
      {:jason, "~> 1.1", only: [:dev, :test], runtume: [:false]},
```

yordis:
Could we improve this PR by moving the `json_library` under a config key and use `Jason` by default.

```

def json_library
end
```

yordis:
Could we improve this PR by moving the json_library under a config key and use Jason by default?

```elixir
  def json_library do
    Application.get_env(:guardian, :json_library, Jason)
  end
```

Then use that function call to get the JSON library

yordis:
@asummers 🎉 🎉 🎉 🎉 🎉 🎉

PR #544, opened 2018-12-02
Fix dialyzer errors and add to build process.

asummers:
`Guardian.t()` wasn't a thing, and I wasn't sure what it was supposed to be, searching the code for similar types. So I just made it say `module`.


asummers:
Just a quick bump on these and my other open ones, seeing if there's anything preventing them from bring merged.

mix.exs
```diff
@@ -163,7 +163,7 @@ defmodule Guardian.Mixfile do
       {:plug, "~> 1.3.3 or ~> 1.4", optional: true},
 
       # Tools
-      {:dialyxir, ">= 0.0.0", only: [:dev], runtime: false},
+      {:dialyxir, ">= 1.0.0-rc4", only: [:dev], runtime: false},
```
yordis: We can't publish any package that depends of a `rc` version.

asummers: This is a local dependency, which would not be published in consumer `mix.lock` files.

yordis: Hex is not aware of that (correct me if I am wrong) so when we try to publish the package it will detect a `rc` package as a dependency so we will be forced to publish our package as `rc` as well.

asummers: Hex is smart enough! 

For example I require rc3 in erlex [here][https://github.com/asummers/erlex/blob/master/mix.exs#L40)

and hex accepts [0.1.6](https://github.com/asummers/erlex/blob/master/mix.exs#L7) as the version without issue.

Makefile
```diff
@@ -4,9 +4,11 @@ deps:
 	mix deps.get
 
 testing: deps
-	mix credo
+	mix do compile --warnings-as-errors --force
 	mix format --check-formatted
+	mix credo
```
doomspork: Could we update this to `credo --strict` please and thank you 😁

.travis.yml
```diff
@@ -12,6 +12,12 @@ otp_release:
 
 script:
   - make testing
+
+cache:
```
doomspork: Holy moly, I did not know this was a thing.  I am going to have to bring this over to other projects.

doomspork:
@yordis I'm _pretty_ sure we can release with a rc package, we just don't want to do that for things like Phoenix where people will expect a new release with every rc.  It's quickly a slippery slop with some dependencies.

.travis.yml
yordis: @doomspork same 😄

Makefile
asummers: Seems like Credo is upset about nested modules in about 10 spots. I can ignore that warning in those places, I can fix, or we can do that in a separate PR. Up to you.

yordis: @doomspork I would prefer to review the `credo --strict` rules fixed in a separate PR.

@asummers probably better to finish with this PR and then make another PR with `--strict` rule turn on and probably will require the fixes to keep the CI happy.

Are you OK with that?

asummers: Fine by me. I'll rebase this and get it up to date with master then I think this should be good to go!

asummers:
Had to change the cache slightly. Now it's super fast.  11-13m down to 1.5m.

asummers:
I have a branch locally branched off this that bumps Credo and fixes the issues, so I can open that pursuant to this being merged.

doomspork:
@asummers are you on Elixir Slack?  We all lurk on #ueberauth if you ever want to come through and discuss potential changes.  

We're always looking to add motivated folks to the team 😁

PR #547, opened 2018-12-03
Add hours to ttl type.

asummers:
`:day/:days` are also valid but were missing from the docs in `guardian.ex`


doomspork:
Thank you @asummers!

PR #554, opened 2018-12-19
Expose public function to allow putting a token into the session

hassox:
Backport the `put_session_token` function into the 0.14.x branch. When you refresh a token there is currently no way to get it into the session to replace the old one :\


hassox:
Yeah we're not going to backport this by default... If folks really need it we can.

Issue #555, opened 2018-12-21
The verify_literal_claims function should be more that an equal check


jsangilve:
While using JWT, there are situations where the same token is valid for different audiences:

```json
{
  "iss": "https://login.example.com/",
  "sub": "custom|user_id",
  "aud": [
    "example_microservices_api",
    "example_customers_api",
    "example_other_api"
  ],
  "iat": 1545404246,
  "exp": 1545411446,
  "azp": "authorized_party_whatever_id",
  "scope": "openid"
}
```

In the example above, the token is valid for 3 different audiences that could be handled by 3 completely different apps. For instance, one application could only require the token to have `example_microservices_api` audience.

Normally, we'd define a Guardian's pipeline using `VerifyHeader` plug as follows:

```elixir
plug Guardian.Plug.VerifyHeader, claims: %{aud:  "example_microservices_api"}
```

**This will fail because the `aud` claim doesn't match the token's `aud` claim** and `Guardian.Plug.Verify.verify_literal_claim` only does an equal check:

https://github.com/ueberauth/guardian/blob/38aa02508fe9e832d7fdd0bb081a4e6f172b7b56/lib/guardian/token/verify.ex#L68-L80.

Could we change this code to something that consider where claim's value is a list?

```elixir
def verify_literal_claims(claims, nil, _opts), do: {:ok, claims}

def verify_literal_claims(claims, claims_to_check, _opts) do
  results = for {k, v} <- claims_to_check, into: [] do
    claims
    |> Map.get(k)
    |> verify_literal_claim(k, v)
  end

  errors = Enum.filter(results, &(elem(&1, 0) == :error))

  if Enum.any?(errors), do: hd(errors), else: {:ok, claims}
end

defp verify_literal_claim(claim_values, key, v) when is_list(claim_values),
  do: if v in claim_values, do: {:ok, claims}, else: {:error, key}

defp verify_literal_claim(claim_value, key, v),
  do: if  claim_value == v, do: {:ok, claims}, else: {:error, key}
```
N.B: I could probably write this code n a better way.

Hanspagh:
I assume this is fixed, so we should close this

PR #558, opened 2018-12-28
Improve Guardian.Token.Verify.verify_literal_claims function

jsangilve:
Solves #555

### DIFF
--- a/lib/guardian/token/verify.ex
+++ b/lib/guardian/token/verify.ex
@@ -44,7 +44,7 @@ defmodule Guardian.Token.Verify do
   configured for the provided module
 
   Allowed drift is measured in seconds and represents the maximum amount
-  of time a token may be expired for an still be considered valid.
+  Gof time a token may be expired for an still be considered valid.
   This is to deal with clock skew.
   """
   def time_within_drift?(mod, time) when is_integer(time) do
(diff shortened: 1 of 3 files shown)


lib/guardian/token/verify.ex
```diff
@@ -62,20 +62,41 @@ defmodule Guardian.Token.Verify do
...
+
+  When the value of a claim is a list, it checks that all values of
+  the same claim in `claims_to_check` are members of the list.
   """
   def verify_literal_claims(claims, nil, _opts), do: {:ok, claims}
 
   def verify_literal_claims(claims, claims_to_check, _opts) do
-    results = for {k, v} <- claims_to_check, into: [], do: verify_literal_claim(claims, k, v)
+    results =
+      for {k, v} <- claims_to_check, into: [] do
```
yordis: Never used `for` on Elixir after 2 years 😄 

Can we use `map` or `reduce` or something similar like that instead of `for`?

lib/guardian/token/verify.ex
```diff
@@ -62,20 +62,41 @@ defmodule Guardian.Token.Verify do
...
+  defp verify_literal_claim(claim_value, value, key) do
+    if valid_claims?(claim_value, value) do
+      {:ok, claim_value}
+    else
+      {:error, key}
+    end
   end
+
+  defp valid_claims?(claim_values, valid) when is_list(claim_values) and is_list(valid),
+    do: Enum.all?(valid, &(&1 in claim_values))
```
yordis: I would prefer to use `do end` over here.

yordis:
@jsangilve amazing ❤️ I left some comments, also the CI is failing.

lib/guardian/token/verify.ex
```diff
@@ -62,20 +62,45 @@ defmodule Guardian.Token.Verify do
...
+        |> Map.get(k)
+        |> verify_literal_claim(v, k)
+        |> case do
+          {:ok, _} -> acc
+          error -> [error | acc]
+        end
+      end)
 
-    errors = Enum.filter(results, &(elem(&1, 0) == :error))
+    if Enum.empty?(errors), do: {:ok, claims}, else: hd(errors)
```
yordis: Could you change it to `do end` block instead of inlining, please?

lib/guardian/token/verify.ex
```diff
@@ -62,20 +62,45 @@ defmodule Guardian.Token.Verify do
...
+
+  When the value of a claim is a list, it checks that all values of
+  the same claim in `claims_to_check` are members of the list.
   """
   def verify_literal_claims(claims, nil, _opts), do: {:ok, claims}
 
   def verify_literal_claims(claims, claims_to_check, _opts) do
-    results = for {k, v} <- claims_to_check, into: [], do: verify_literal_claim(claims, k, v)
+    errors =
+      claims_to_check
```
yordis: Don't use pipe operator when there is only one function call

lib/guardian/token/verify.ex
```diff
@@ -62,20 +62,45 @@ defmodule Guardian.Token.Verify do
...
+      |> Enum.reduce([], fn {k, v}, acc ->
+        claims
```
yordis: Try to extract the block before the `case` into a function so this `reduce` block is pretty much the `case` statement.

jsangilve:
@yordis, Do you guys want something else here or is it ready to be merged? Thx

lib/guardian/token/verify.ex
```diff
@@ -44,7 +44,7 @@ defmodule Guardian.Token.Verify do
   configured for the provided module
 
   Allowed drift is measured in seconds and represents the maximum amount
-  of time a token may be expired for an still be considered valid.
+  Gof time a token may be expired for an still be considered valid.
```
doomspork: @jsangilve could you remove this `G` that slipped in? Otherwise this is good IMO.

PR #560, opened 2019-01-14
Fix links and formatting

toraritte:
Did a minor re-phrasing of the first paragraph at `## Pipelines`.


doomspork:
@yordis / @toraritte should we add a note to the README about generating the documentation for future users?

yordis:
@doomspork I would, I can't count how many times I have to explain this 😄 , but I would say that maybe Elixir itself should invest a little bit more in clarifying these things, but we do not have control of it.

PR #564, opened 2019-01-16
V2

yordis:
## Breaking Changes

- Remove Phoenix socket functionality, it is moved to `guardian_phoenix` package


doomspork:
I'm definitely onboard with pulling out Phoenix because it includes far more than needed for Guardian to function. I'm not entirely sure the same can be said for Plug.  

So my first question is by pulling out Plug what we do hope to gain?

My concern remains: if 99% of Guardian users are using it with Plug in some fashion (including Phoenix) pulling it out of `guardian` just makes using the library involve that many additional steps. I'm also not convinced Plug so large a dependency that even if you weren't using Guardian for that, you're not pulling in dozens of sub-dependencies resulting in some insane dependency tree.

@hassox / @scrogson any thoughts on pulling out Plug vs keeping it?

Hanspagh:
Seems like pulling out phoenix is reasonable, but I still think there is some value in the helper functions in [socket](https://github.com/ueberauth/guardian/blob/master/lib/guardian/phoenix/socket.ex). Would it somehow make sense to keep these to avoid having people using the sockets, implement this them self? I am not sure how this would be achieved though :confused:

config/config.exs
```diff
@@ -1,26 +1,3 @@
-# This file is responsible for configuring your application
-# and its dependencies with the aid of the Mix.Config module.
```
Hanspagh: Any reason to remove these two lines of documentation?

doomspork:
@Hanspagh I'm not familiar with using sockets outside of Phoenix, when/why would you need that?

.travis.yml
```diff
@@ -4,11 +4,12 @@ notifications:
   email: false
 
 elixir:
-  - 1.6
+  - 1.7
+  - 1.8
```
doomspork: We need to add 1.9 now, right?

yordis: @doomspork is there a version `1.9` released? I wasn't aware of it.

doomspork: I'm confused, you're right 👍

config/config.exs
yordis: There are two reasons.

1. This is a misleading message for many new programmers in Elixir land; this does not configure your application if you use this package as a dependency, has been multiple (no on this package) cases where they assume they need to add some default values to these configs.

2. Having those messages add no value to the source code, this is well documented in Elixir for us to have these messages all around the repositories.

In any case, feel free to block the merging, and I will roll the changes back if you think this is not a good idea.

Hanspagh:
@doomspork I was still thinking sockets within phoenix, but my thought was that everybody would need the helper functions when working with the phoenix sockets, so it would be nice to have them somewhere, maybe in a guide or a utility library.

doomspork:
@Hanspagh all of the Phoenix functionality is moving into `guardian_phoenix`, so folks will still have access to them. We'll update the guides to reflect the new package.

Hanspagh:
@doomspork I completely missed that, my apologies. The above can just be ignored 😊

doomspork:
No worries @Hanspagh! Glad we're on the same page 😁

PR #567, opened 2019-01-19
Expand Travis config

doomspork:


yordis:
@doomspork don't we agree on be responsible only for the last 2 versions of Elixir?

In this case, `1.7` and `1.8`

doomspork:
Removed

yordis:
It seems that there is a mismatch betwen Elixir version and supported OTP version

yordis:
I would say remove the matrix and do the following config

```
elixir:
  - 1.7
  - 1.8

otp_release:
  - 21.1
  - 21.2
```

doomspork:
@yordis for my own edification can you explain how you figured that out? Reading the Travis logs it looks like Dialyzer took too long. Each of the apps built just fine.

PR #570, opened 2019-01-24
Override cookie defaults in remember_me with values from module config

sparta-developers:
We found that when we attempt to set the `max_age` in our Guardian module's config, this value would not override the default `cookie_options` in `Guardian.Plug.remember_me`. This is because the options list is concatenated (defaults last) and eventually Phoenix uses `:maps.from_list/1` which prefers the right most value. This PR should fix the issue. 

Example config:
```
  issuer: "MyApp",
  secret_key: "...",
  permissions: %{
    default: [:read_users, :write_users]
  },
  cookie_options: [max_age: 100]
```

### DIFF
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -291,7 +291,7 @@ if Code.ensure_loaded?(Plug) do
     end
 
     defp cookie_options(mod, _claims) do
-      mod.config(:cookie_options, []) ++ @default_cookie_options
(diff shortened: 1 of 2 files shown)


yordis:
@sparta-developers hey, would you mind to fix the issues from the Travis? You need to run `mix format`

sparta-developers:
@yordis fixed. Thanks

PR #571, opened 2019-01-25
Fix dialyzer warning in Guardian.Plug

barthez:
Hey,

I noticed that dialyzer complains about supertype of success specification of generated `implementation/0` function in generated `(...).Plug` module:

```
Type specification 'Elixir.MyApp.Guardian.Plug':implementation() -> module() is a supertype of the success typing: 'Elixir.MyApp.Guardian.Plug':implementation() -> 'Elixir.MyApp.Guardian'
```

There is only 1 known value returned from that function, so we can also put it in function specification.

### DIFF
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -54,7 +54,7 @@ if Code.ensure_loaded?(Plug) do
 
     defmacro __using__(impl) do
       quote do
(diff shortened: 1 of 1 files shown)


doomspork:
Sorry @barthez, I missed your PR and fixed this via another merge.

PR #580, opened 2019-02-26
Update getting started tutorial to phoenix 1.4

LouisDelbosc:
I went through the getting started with the last phoenix version and got into some minor troubles.
Here the code modification I had to operate for my code to work.

Thank you for your work :)


Hanspagh:
Hi, would you mind telling what kind of trouble you had? The changes are fine, but ` {:bcrypt_elixir, "~> 2.0"}` requires Erlang 20, which should be okay, but I would still be happy to know what was not working.

PR #583, opened 2019-03-09
Update deps

yordis:


doomspork:
Tests are failing @yordis

yordis:
@doomspork it seems to be Phoenix deps, probably I will hold this PR until #564 is merged

doomspork:
@yordis would we want want this PR and #584 as part of V2?

yordis:
yeah much better actually

Issue #578, opened 2019-02-19
Improve the error messages of the permission plug 


Hanspagh:
Right now the error message is always `{:unauthorized, :unauthorized}` which is not really helping so much when trying to figure out why a pipeline failed the authentication.

yordis:
@Hanspagh is this fixed?

Hanspagh:
No, we still have the {:unauthorized, :unauthorized} [here](https://github.com/ueberauth/guardian/blob/master/lib/guardian/permissions/permissions.ex#L430). I will find some time to work on it

yordis:
@Hanspagh I remember to be working on this thou, I have no clue what happened.

Let me get back to it then.

yordis:
LOL I found it https://github.com/ueberauth/guardian/pull/589

PR #584, opened 2019-03-09
Fix error messages

yordis:
Related to #578

@ueberauth/developers I need your feedback on this, we currently send an error tuple to the clients with a text body containing that error tuple.

Should I send a human-readable message in this case?

I can fix this but I need your direction on what the proper fix for this is.

### DIFF
--- a/test/guardian/permissions/bitwise_test.exs
+++ b/test/guardian/permissions/bitwise_test.exs
@@ -292,7 +292,7 @@ defmodule Guardian.Permissions.BitwiseTest do
 
       assert conn.halted
       assert {403, _headers, body} = sent_resp(conn)
-      assert body == "{:unauthorized, :unauthorized}"
+      assert body == "{:unauthorized, :missing_claims}"
     end
 
     test "when looking in a different location with correct permissions", ctx do
--- a/test/guardian/permissions/bitwise_test.exs
+++ b/test/guardian/permissions/bitwise_test.exs
@@ -224,7 +224,7 @@ defmodule Guardian.Permissions.BitwiseTest do
       conn = Guardian.Permissions.Bitwise.call(ctx.conn, opts)
 
       assert {403, _headers, body} = sent_resp(conn)
-      assert body == "{:unauthorized, :unauthorized}"
+      assert body == "{:unauthorized, \"Insufficient permission\"}"
       assert conn.halted
     end
 
@@ -240,7 +240,7 @@ defmodule Guardian.Permissions.BitwiseTest do
       conn = Guardian.Permissions.Bitwise.call(ctx.conn, opts)
 
       assert {403, _headers, body} = sent_resp(conn)
-      assert body == "{:unauthorized, :unauthorized}"
+      assert body == "{:unauthorized, \"Insufficient permission\"}"
       assert conn.halted
     end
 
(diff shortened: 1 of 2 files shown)


doomspork:
I see no point in returning a tuple as a string. We should probably just be sending the `reason`.

yordis:
@doomspork actually we call `auth_error` from `AuthErrorHandler` module so, should I send an `atom` instead of a `string`?

test/guardian/permissions/bitwise_test.exs
```diff
@@ -224,7 +224,7 @@ defmodule Guardian.Permissions.BitwiseTest do
       conn = Guardian.Permissions.Bitwise.call(ctx.conn, opts)
 
       assert {403, _headers, body} = sent_resp(conn)
-      assert body == "{:unauthorized, :unauthorized}"
+      assert body == "{:unauthorized, \"Insufficient permission\"}"
```
doomspork: Don't we want to return just "Insufficient permission" here?

yordis: @doomspork welcome to feedbacks

yordis: I am not sure what would people prefer to get here 🤷‍♂️

doomspork: The including the tuple seems unnecessary don't you think? Tuples nor atoms are valid JSON so this isn't anything, right? It's not real sentence and it's not valid JSON.

yordis: > The including the tuple seems unnecessary don't you think? Tuples nor atoms are valid JSON so this isn't anything, right?

Didn't write this so I am not sure, never used `bitwise` module before and the reason I asked for help.

But from what I understand this is being passed to the `auth_error` callback where you decide to respond with a valid JSON, text, or whatever you decide to do.

Hanspagh: I used it when I was debugging a bitwise error. What I thought could be cool was to write the input permission and the required permission, but maybe this is a little bit out of scope for now?

doomspork: @yordis I understand but we should improve the code anytime we make changes to it 😁In addition to testing the code, the tests can/should service has a type of documentation. Never would you really return a tuple like this, so let's change it, even if you didn't write it.

If you come to Guardian to either use the library or contribute, find your way to these tests, and see them returning something that _you'd never see in an application_ it could be a source of confusion — hell it's confused me and I'm one of the maintainers 😆

Let's update the test handler to return something like valid JSON so these tests can look remotely like they would in a real scenario.

Does that sounds good?

```elixir
  defmodule Handler do
    @moduledoc false

    import Plug.Conn
    @behaviour Guardian.Plug.ErrorHandler

    @impl Guardian.Plug.ErrorHandler
    def auth_error(conn, {type, reason}, _opts) do
      send_resp(conn, 403, Poison.encode!(%{type => reason}))
    end
  end```

yordis: @doomspork since users will handle what to do with the error tuple anyway I will put as many details as possible on what happened (like @Hanspagh wants it), in any case, the user could decide if they want to send a generic error or not (this is why I put a generic error, I wasn't sure about it).

Now, if that is the case, maybe instead of returning a string back I will return an atom since the user can always decide to convert the atom to the proper string message.

Or triple tuple with the last value being the human-readable message (I would simply work with atoms).

doomspork: @yordis I'm only talking about updating this test. I'm not sure I follow all these other changes?

doomspork: https://github.com/ueberauth/guardian/blob/77c7acdba384563160b935bf75d233bf71f495b7/test/guardian/permissions/bitwise_test.exs#L27

yordis:
@doomspork @Hanspagh let's do this on #585

Issue #470, opened 2018-02-21
Option for encoding permissions as text instead of bitstrings


jgashe:
For clients that aren't elixir applications that use Guardian, decoding permissions as bitstrings adds unnecessary hassle. Cowboy's header length limit is 4k bytes (as is NGINX), which is plenty of room to encode permissions as a string array of the granted permissions. 

I propose a config setting `permission_encoding` with values `:bitstring` and `:text` (default `:bitstring`) to determine which strategy to use.

hassox:
Hey @jgashe 
Thanks for the thought. If love to see an option to encode as string front is one, but I don't think it should be an option on the bitwise module. I was thinking if we want a different kind we should make a different module (hence the btwise in it's name). On the other hand we could pull out the encoding/decoding part into a behaviour interface and have a bitstring, text, JSON encoders.

Thoughts

mikemorris:
> On the other hand we could pull out the encoding/decoding part into a behaviour interface and have a bitstring, text, JSON encoders.

This sounds like a great proposal, I'd love to see an interface similar to https://www.npmjs.com/package/express-jwt-permissions supported (space-delimited string and/or array of strings).

Hanspagh:
This is gonna be part of V2

PR #585, opened 2019-03-12
New permission types

Hanspagh:
I have started to refactor the bitwise permission into its own module. This will to allow other encodings like json, as suggest here #470

Currently it is very much work in progress, but as of now I have moved the bitwise encoding to its own module and all the old tests parses. Next task is to implement a json encoder and then make the interface simpler to make it easier for people to write their own encoders. 

To avoid breaking changes I didn't change the module name from bitwise to permissions, but maybe it is something we should consider?


doomspork:
@Hanspagh I will look through this later today or tomorrow but I'm loving the refactoring so far! I'm all for improving the maintainability of the code. 🎉

lib/guardian/permissions/bitwise_encoding.ex
```diff
@@ -0,0 +1,28 @@
...
+    value
+  end
+
+  def encode(value, type, perm_set) when is_list(value) do
+    perms = Map.get(perm_set, type)
+    Enum.reduce(value, 0, &encode_value(&1, perms, &2))
+  end
+
+  defp encode_value(value, perm_set, acc),
+    do: perm_set |> Map.get(to_string(value)) |> bor(acc)
```
yordis: I would prefer to have `do ... end` since it makes it harder to read, no a big deal whichever way.

doomspork:
@Hanspagh now would be the time to change the module name. While V2 is merged, it has not been released, and this could be included as V2 is a breaking change.

Thoughts?

cc @yordis

yordis:
Definitely, prefer to do the changes that would introduce breaking change as V2

Hanspagh:
I have now renamed the base permission module and added text and atom encoding modules. Fell free to have a look.

Hanspagh:
Any feedback on this?

yordis:
@Hanspagh sorry, I wasn't aware if this was ready for code review or not. I will do some CR today.

doomspork:
Sorry @Hanspagh, like @yordis I saw the WIP I thought this was still being worked on. I removed the label and I'll look this over today.

lib/guardian/permissions/text_encoding.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Guardian.Permissions.TextEncoding do
+  @behaviour Guardian.Permissions.PermissionEncoding
+  use Bitwise
+
+  def encode(value, type, perm_set) when is_integer(value) do
+    perms = Map.get(perm_set, type)
+
+    for {k, v} <- perms, band(value, v) == v, into: [] do
+      k |> to_string()
```
doomspork: Please don't use pipelines for a single function call, this can be `to_string(k)` 👍

lib/guardian/permissions/text_encoding.ex
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  defp encode_value(value, _perm_set, acc) when is_atom(value),
+    do: [Atom.to_string(value) | acc]
+
+  defp encode_value(value, _perm_set, acc) when is_binary(value),
+    do: [value | acc]
+
+  def decode(value, _type, _perm_set) when is_list(value) do
+    value |> Enum.map(&String.to_atom/1)
```
doomspork: `Enum.map(value, &String.to_atom/1)`

yordis:
@doomspork should we use `do ... end` block over `do: ` all the time?

This is the only thing that I would give feedbacks so far but I am not sure if that should be the case.

doomspork:
@yordis only if it is multiple lines. I think trying to use `, do:` with long single-line functions makes it difficult to read.

Hanspagh:
I sorry for the long delay, I will fix the last requested changes today

Hanspagh:
Fixed the pipelines issues, bad habit of just putting everything in a pipe

yordis:
@Hanspagh travis is failing btw

Hanspagh:
Travis should be happy now

doomspork:
Thank you @Hanspagh!!

doomspork:
@yordis would you like to prep the next release?

yordis:
@doomspork Aye!

Then I need to close https://github.com/ueberauth/guardian_phoenix/issues/2

PR #586, opened 2019-04-11
Remove `!` from UserManager.get_user!

blisscs:
Remove `!` from UserManager.get_user!, since we are pattern matching with nil in the case block.


yordis:
🚀

PR #588, opened 2019-05-22
Update changelog

yordis:


yordis:
@ueberauth/developers can you share some love with CR

PR #590, opened 2019-05-23
Improve Permissions

doomspork:


lib/guardian/permissions.ex
```diff
@@ -375,96 +369,4 @@ defmodule Guardian.Permissions.Permissions do
       {String.to_atom(k), list}
     end
   end
-
-  if Code.ensure_loaded?(Plug) do
```
doomspork: I pulled this all out into another module so it's easier to manage.

PR #592, opened 2019-06-03
Fix controller example. Change route resource name.

moxley:
Made some fixes to the SessionController example. The calls to `Guardian.Plug.*` didn't have the correct arguments.

Renamed the "secret" resource to "protected", to clarify the meaning.

Changed the "logout" endpoint to be an http `GET`. While `POST` may be technically more correct, `GET` is easier to implement.

Changed the Phoenix flash type from `:success` to `:info`, because `:info` is one of the types that come pre-configured.


yordis:
@moxley 🚀

PR #596, opened 2019-07-03
Change `clear_remember_me` to put an immediately expiring cookie

### LINKED ISSUE #595: No way to delete `remember_me` cookie from client
After I've done `remember_me_from_token(conn, token, claims)` in the `login` controller method, I'd like to do something in the `logout` method so that a `set-cookie` header with the token cookie removed is sent to the client.

Currently my best bet is to copy the private [`fetch_token_key`](https://github.com/ueberauth/guardian/blob/9ed23acbef1ddfde5a79c139570fada9399b0877/lib/guardian/plug.ex#L281) to my code and use that together with `put_resp_cookie` to manually clear the cookie.

EDIT: this worked (in my controller):

```elixir
  def logout(conn, _) do
    key = fetch_token_key(conn)

    conn
    |> Guardian.Plug.sign_out()
    |> Guardian.Plug.clear_remember_me()
    |> put_resp_cookie(key, "deleted", Accounts.Auth.Guardian.config(:cookie_options) ++ [ max_age: 0 ])
    |> render("logout.json")
  end

  # Workaround for https://github.com/ueberauth/guardian/issues/595
  defp fetch_token_key(conn, opts \\ []) do
    conn
    |> Pipeline.fetch_key(opts)
    |> Keys.token_key()
    |> Atom.to_string()
  end
```

futpib:
(instead of just removing an existing one from the response)

Fixes #595

### DIFF
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -95,7 +95,7 @@ if Code.ensure_loaded?(Plug) do
           do: Guardian.Plug.remember_me_from_token(conn, implementation(), token, claims, opts)
 
         def clear_remember_me(conn, opts \\ []),
-          do: Guardian.Plug.clear_remember_me(conn, opts)
+          do: Guardian.Plug.clear_remember_me(conn, implementation(), opts)
       end
     end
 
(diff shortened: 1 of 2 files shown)


Hanspagh:
Thank you for work @futpib, looks good to me. This should not be a breaking change since the old implementation assumed the user would clear the cookie them self

PR #597, opened 2019-07-09
Allow ability to verify token in custom header location

vaer-k:
Currently, tokens can only be verified under the "authorization" header, and only the "realm" can be customized with options on the plug. This change will allow users to verify tokens in headers other than those found in the "authorization" field, like so:

```
plug Guardian.Plug.VerifyHeader,
  header_name: "x-access-token",
  realm: :none,
  claims: %{"typ" => "access"}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Allow ability to verify token in custom header location
+
 # v 1.1.0
 
 * JWT secret fetcher behaviour added
--- a/lib/guardian/plug/verify_header.ex
+++ b/lib/guardian/plug/verify_header.ex
@@ -30,7 +30,8 @@ if Code.ensure_loaded?(Plug) do
     Options:
 
     * `claims` - The literal claims to check to ensure that a token is valid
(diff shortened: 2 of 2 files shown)


yordis:
@vaer-k could you add some documentation about this as well?

vaer-k:
> @vaer-k could you add some documentation about this as well?

Of course; happy to.

yordis:
@vaer-k also I forgot, can you add an entry to `CHANGELOG.md` describing the change for the incoming releases, please.

vaer-k:
@yordis What version number should this go under?

yordis:
@vaer-k put it at the top since we didn't create a release for

vaer-k:
@yordis OK, all set.

yordis:
@vaer-k I am sorry I meant before line `1` like the first line since this doesn't belong to that release, sorry for the confusion.

vaer-k:
@yordis No worries, I'm just not clear what version is current so it's not clear whether 1.1 is released yet or not. Anyway thanks for your guidance. I've placed this update on the top of the file with no version header above it, so that you can add that line when you release whatever version is next. I hope that's what you wanted?

Issue #599, opened 2019-07-12
Issuing different jwt tokesn not possible


Sgoettschkes:
I'm trying to use guardian to generate multiple jwt tokens. The use case is that we have an existing authentication provider we want to deprecate so the new one should generate a new jwt token and a legacy one which will be removed after all applications are using the new token.

The new token will be using "RS512" with a public/private key and it's working as expected. The legacy token uses "HS256". This is the code:

```
AuthenticationService.Guardian.encode_and_sign(user, %{},
  ttl: {42, :minutes},
  secret_fetcher: Guardian.Token.Jwt.SecretFetcher,
  secret: "secret",
  allowed_algos: ["HS256"]
)
```

As shown, I'm overwriting the fetcher to not use the public/private key fetcher, use our old secret and also changed the allowed algorithm.

I'm getting the the error`(ErlangError) Erlang error: {:not_supported, [:HS256]}` and it seems that the optional arguments are overwritten by config values.

I'm under the impression that I'm doing something wrong but I can't seem to figure it out!

Sgoettschkes:
As a follow up, if I exchange the config to support the "HS256" algorithm, everything works fine. I looked through the source and from what I can see the `fetch_secret_fetcher` function 
https://github.com/ueberauth/guardian/blob/5663cfa94bdabe33474accb7b49052ca12ea4ea5/lib/guardian/token/jwt.ex#L510-L512
only reads from the config, never from the opts, ignoring the option.

Could this be the issue?

yordis:
@Sgoettschkes make a PR with your proposal on how to fix the current issue. it will help us to understand exactly what needs to change.

Hanspagh:
@Sgoettschkes I am not sure it was ever the intention to be able to pass the secret_fetcher as a optional param to `decode_and_sign`. I was thinking that creating two Guardian config would fix this issue for you.
Basicly would have one config what is your old token issuer and one for the new private public key one. Having them in two modules also makes it a bit more easy to maintain and document. What do you think?

Sgoettschkes:
Maybe I misunderstood the docs. From the docblocks I got 

> `opts` - The options to pass to the token module and callbacks

and assumed you could overwrite the config options.

I can see how two different configs would solve my problem and will try that. Would you be interested in me adding a few sentences of documentation for this use case?

Hanspagh:
Yea, I can see why I can be a bit misleading. Feel very welcome contribute
to make it more clear

[...]

PR #601, opened 2019-07-15
Implement using option `secret_fetcher`

Sgoettschkes:
As discussed in #599, currently the `secret_fetcher` option which can be passed to `Guardian.decode_and_sign` isn't taken into account when choosing the secret fetcher.

I added a test to verify this bug and implemented the reading of this option value in the `fetch_secret_fetcher` function.

ToDos:

- [ ] Update documentation (docstrings) to reflect new option for JWT functions
- [ ] Add test to verify correct behaviour on decode
- [ ] Format code so satisfy Travis CI build

I'd be happy to review any feedback and suggestions for improvement.

### DIFF
--- a/lib/guardian/token/jwt.ex
+++ b/lib/guardian/token/jwt.ex
@@ -248,7 +248,7 @@ defmodule Guardian.Token.Jwt do
 
   """
   def create_token(mod, claims, options \\ []) do
-    with {:ok, secret_fetcher} <- fetch_secret_fetcher(mod),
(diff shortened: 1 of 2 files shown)


Sgoettschkes:
Before I tackle the Todos, could a contributor verify that I'm on the right path and that it's worth pursuing?

Hanspagh:
Since this is better fixed by using different configs, I am closing this for now

PR #602, opened 2019-07-16
Tutorial: Fixed major errors, minor errors.

### LINKED ISSUE #600: UserManager.change_user() ?
(1)
The only function that the tutorial puts in `lib/auth_me/user_manager.ex` is:

```
  def authenticate_user(username, plain_text_password) do
    query = from u in User, where: u.username == ^username
    case @repo.one(query) do
      nil ->
        Argon2.no_user_verify()
        {:error, :invalid_credentials}
      user ->
        if Argon2.verify_pass(plain_text_password, user.password) do
          {:ok, user}
        else
          {:error, :invalid_credentials}
        end
    end
  end

```

Yet, in  `lib/auth_me_web/controllers/session_controller.ex` the tutorial calls the function:

```
    UserManager.change_user(%User{})
```
here:
```
defmodule AuthMeWeb.SessionController do
  use AuthMeWeb, :controller

  alias AuthMe.{UserManager, UserManager.User, UserManager.Guardian}

  def new(conn, _) do
    changeset = UserManager.change_user(%User{})  ### <===== HERE 
```

Is `AuthMe.UserManager.change_user()` simply a wrapper around `AuthMe.UserManager.User.changeset()`, for instance:

```
def change_user(user, attrs \\ %{}), do: AuthMe.UserManager.User.changeset(user, attrs)
```

(2)  
In `lib/auth_me/user_manager.ex`:

```
  def authenticate_user(username, plain_text_password) do
    query = from u in User, where: u.username == ^username
```

the `query` line will throw an error without:

```
 import Ecto.Query, only: [from: 2]
```

(3) 
The tutorial calls another undefined function in the [Try it Out](https://github.com/ueberauth/guardian/blob/master/guides/tutorial/start-tutorial.md#try-it-out) section:
```
Create the user:
AuthMe.UserManager.create_user(%{username: "me", password: "secret"})
```

7stud:
closes #600 

1) Because the tutorial says that you can use your own User module instead of doing:
```
$ mix phx.gen.context UserManager User users username:string password:string
```

it should be made explicit that if you do use your own User module, then you need to define certain functions for the tutorial to work properly.

2) In the file `lib/auth_me/user_manager/user_manager.ex`, the tutorial doesn't specify the required import statement to allow the query to work without error:
```
defmodule AuthMe.UserManager do
  ...
  ...
  def authenticate_user(username, plain_text_password) do
    query = from u in User, where: u.username == ^username
    ...
    ...
```
The required import statement is:
```
  import Ecto.Query, only: [from: 2]
```

3) The tutorial alternately uses `"/secret"` and `"/protected"` for the route to the protected page.  The route needs to be the same everywhere in the tutorial.

4)  As currently written, the function `AuthMeWeb.SessionController.login_reply()` (in the file `lib/auth_me_web/controllers/session_controller.ex`) will result in the error:

> protocol Enumerable not implemented for %AuthMe.UserManager.User{meta: 
> #Ecto.Schema.Metadata<:loaded, “users”>, id: 3, inserted_at: ~N[2019-07-15 05:56:22], password: 
>“$argon2id$v=19$m=131072,t=8,p=4$jinhp9gEDVENO2PIdm5zkg$ovideTURD59gCoqtE6UUuYZ/32Y
> MnpgppuRf0NFAJ/k”, updated_at: ~N[2019-07-15 05:56:22], username: “me”}. This protocol is
>  implemented for: Ecto.Adapters.SQL.Stream, Postgrex.Stream, DBConnection.Stream, 
> DBConnection.PrepareStream, HashSet, Range, Map, Function, List, Stream, Date.Range, HashDict, 
> GenEvent.Stream, MapSet, File.Stream, IO.Stream
> 

Here's the offending code:
```
defmodule AuthMeWeb.SessionController do
  ...
  alias AuthMe.{UserManager, UserManager.User, UserManager.Guardian}
  ...
  ...
  defp login_reply({:ok, user}, conn) do
    conn
    |> put_flash(:info, "Welcome back!")
    |> Guardian.Plug.sign_in(Guardian, user)
    |> redirect(to: "/secret")
  end
```
Due to the alias, the line:
```
Guardian.Plug.sign_in(Guardian, user)
```

is equivalent to:
```
AuthMe.UserManager.Guardian.Plug.sign_in(Guardian, user)
```

And, because a connection is piped in as the first argument, the call becomes:
```
AuthMe.UserManager.Guardian.Plug.sign_in(connection, Guardian, user)
```

If you refer to the `Guardian.Plug.sign_in()` [docs ](https://hexdocs.pm/guardian/Guardian.Plug.html#sign_in/5), the arguments are correct.  However, the definition of the function:
```
AuthMe.UserManager.Guardian.Plug.sign_in()
```
in the guardian source code is different than the definition in docs for the function:
```
Guardian.Plug.sign_in() 
```

How is that possible?  The function `AuthMe.UserManager.Guardian.Plug.sign_in()` is injected into the file `/lib/auth_me/user_manager/guardian.ex` by the statement:
```
use Guardian, otp_app: :auth_me
```

which calls the `Guardian.__using__()` macro. The `Guardian.__using__()` macro is defined in the source code [here](https://github.com/ueberauth/guardian/blob/master/lib/guardian.ex#L322) and contains the line:
```
use Guardian.Plug, unquote(__MODULE__)
```

which calls the `Guardian.Plug.__using__()` macro.  The  `Guardian.Plug.__using__()` macro is defined in the source code [here](https://github.com/ueberauth/guardian/blob/dfb77728a44c35b8b929a874a41d8e20aa0aa69e/lib/guardian/plug.ex#L55), and it contains the code:
```
quote do
    ...
    ...
    def sign_in(conn, resource, claims \\ %{}, opts \\ []),
          do: Guardian.Plug.sign_in(conn, implementation(), resource, claims, opts)
```
As can be see in that definition for `sign_in()`, the function should be called with two arguments: `conn` and the resource, i.e. `user`, not: `conn, Guardian, user`.  When `sign_in()` is called with the latter arguments, `user` gets assigned to the parameter variable `claims`, and elsewhere in the guardian.ex source code the following executes:
```
      claims
      |> Enum.into(%{})
```
which causes the error: `protocol Enumerable not implemented for %AuthMe.UserManager.User`

4) A similar problem as in 3) exists for the logout() function in AuthMeWeb.SessionController.

5) At the end of the tutorial, I changed the description of how a user can verify that Guardian is actually working.  The new description instructs the user to try to gain access to the "/protected" route without first logging in, as well as steps to experiment with the logout feature. 

6) Fixed other minor errors.

### DIFF
--- a/guides/tutorial/start-tutorial.md
+++ b/guides/tutorial/start-tutorial.md
@@ -37,10 +37,45 @@ end
 
 ## Create a user manager
 
-We'll need something to authenticate. How Users are created and what they can do is outside the scope of this tutorial. If you already have a user model you can skip this part.
+We'll need something to authenticate. How Users are created and what they can do is outside the scope of this tutorial. If you already have a User model you can skip this part (but see the note after the command).
 
 ```sh
 $ mix phx.gen.context UserManager User users username:string password:string
+```
+
+Note: if you already have a User model, you need to implement some of the
+functions that mix generates for the UserManager context. mix generates the
+following functions:
+
+```elixir
+
+  def list_users do
+    Repo.all(User)
+  end
+
+  def get_user!(id), do: Repo.get!(User, id)
+
+  #The tutorial calls this one:
+  def create_user(attrs \\ %{}) do
+    %User{}
+    |> User.changeset(attrs)
+    |> Repo.insert()
+  end
+
+  def update_user(%User{} = user, attrs) do
+    user
+    |> User.changeset(attrs)
+    |> Repo.update()
+  end
+
+  def delete_user(%User{} = user) do
+    Repo.delete(user)
+  end
+
+  #The tutorial calls this one:
+  def change_user(%User{} = user) do
+    User.changeset(user, %{})
+  end
 ```
 
 ## Create implementation module
@@ -139,6 +174,8 @@ Now we need a way to verify the username/password credentials.
 ## lib/auth_me/user_manager/user_manager.ex
 
 alias Argon2
+import Ecto.Query, only: [from: 2]
+
 
 def authenticate_user(username, plain_text_password) do
   query = from u in User, where: u.username == ^username
@@ -249,7 +286,7 @@ defmodule AuthMeWeb.SessionController do
     changeset = UserManager.change_user(%User{})
     maybe_user = Guardian.Plug.current_resource(conn)
     if maybe_user do
-      redirect(conn, to: "/secret")
+      redirect(conn, to: "/protected")
     else
       render(conn, "new.html", changeset: changeset, action: Routes.session_path(conn, :login))
     end
@@ -262,17 +299,17 @@ defmodule AuthMeWeb.SessionController do
 
   def logout(conn, _) do
     conn
-    |> Guardian.Plug.sign_out(Guardian, _opts = [])
-    |> redirect(to: "/login")
-  end
+    |> Guardian.Plug.sign_out() #This module's full name is Auth.UserManager.Guardian.Plug, 
+    |> redirect(to: "/login")   #and the arguments specfied in the Guardian.Plug.sign_out()
+  end                           #docs are not applicable here               
 
   defp login_reply({:ok, user}, conn) do
     conn
     |> put_flash(:info, "Welcome back!")
-    |> Guardian.Plug.sign_in(Guardian, user)
-    |> redirect(to: "/secret")
-  end
-
+    |> Guardian.Plug.sign_in(user)   #This module's full name is Auth.UserManager.Guardian.Plug,
+    |> redirect(to: "/protected")    #and the arguments specified in the Guardian.Plug.sign_in() 
+  end                                #docs are not applicable here.                 
+                                               
   defp login_reply({:error, reason}, conn) do
     conn
     |> put_flash(:error, to_string(reason))
@@ -344,7 +381,7 @@ We use the `Guardian.Plug.current_resource(conn)` function here to fetch the res
 
 ## Routes
 
-Ok. So the controller and views are not strictly part of Guardian but we need some way to interact with it. From here the only thing left for us to do is to wire up our router.
+Okay. So the controller and views are not strictly part of Guardian but we need some way to interact with it. From here the only thing left for us to do is to wire up our router.
 
 ```elixir
 # Our pipeline implements "maybe" authenticated. We'll use the `:ensure_auth` below for when we need to make sure someone is logged in.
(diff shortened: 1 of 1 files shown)


guides/tutorial/start-tutorial.md
```diff
@@ -37,12 +37,47 @@ end
...
 
-We'll need something to authenticate. How Users are created and what they can do is outside the scope of this tutorial. If you already have a user model you can skip this part.
+We'll need something to authenticate. How Users are created and what they can do is outside the scope of this tutorial. If you already have a User model you can skip this part (but see the note after the command).
 
 ```sh
 $ mix phx.gen.context UserManager User users username:string password:string
 ```
 
+Note: if you already have a User model, you need to implement some of the
+functions that mix generates for the UserManager context. mix generates the
```
Hanspagh: I think practically its phoenix that generates these :)

7stud: Okay.  I wanted to make sure the reader knew exactly what command I was referring to.   How about I change `mix` to phx.gen.context`?  And...how do I do that?  Do I make a new pull request with another branch name with just that change?

Hanspagh: I completely agree that `phx.gen.context` is better. To fix it simply just make a new commit on your current branch and push it. Then it will show up here and will do a Squash and merge to make it into one commit in the end :)

7stud: Okay, I'll do that.  I was reviewing some of the text and doing some additional editing, and I occasionally found it confusing  when the tutorial refers to a `resource`.  In Guardian speak, a resource is a User.  However, the tutorial also uses the word resource in the web sense of the word, e.g. a web page.   I thought it would be clearer if I changed the word `resource` to `User` when appropriate, and to `web page` when that was the intended meaning.   But, in the bolded text below I cannot figure out in what sense the word resource is being used:  

> Routes
> 
> Okay. So the controller and views are not strictly part of Guardian but we need some way to interact with Guardian. From here, the only thing left for us to do is to wire up our router.
> 
> # Our AuthMeUserManager.Pipeline module implements "maybe" authenticated. We'll use an `:ensure_auth` pipeline below for when we need to make sure someone is logged in.
> ```
> pipeline :auth do
>   plug AuthMe.UserManager.Pipeline
> end
> 
> # The ensure_auth pipeline will fail if no one is logged in:
> pipeline :ensure_auth do
>   plug Guardian.Plug.EnsureAuthenticated
> end
> 
> # Maybe logged in routes:
> scope "/", AuthMeWeb do
>   pipe_through [:browser, :auth]
> 
>   get "/", PageController, :index
> 
>   get "/login", SessionController, :new
>   post "/login", SessionController, :login
>   get "/logout", SessionController, :logout
> end
> 
> # Definitely logged in routes:
> scope "/", AuthMeWeb do
>   pipe_through [:browser, :auth, :ensure_auth]
> 
>   get "/protected", PageController, :protected
> end
> 
> ```
> There's a little bit happening here:
> 
> We created a Phoenix pipeline that just delegates to our Guardian pipeline to login someone if we find a token in the session info or Authorization header. This does not restrict access.
> We decide what type of access the User has with the :auth and :ensure_auth phoenix pipelines. In order to protect our protected route, we use :ensure_auth.
> Note that you must use the :auth pipeline before the :ensure_auth pipeline to make sure that we have fetched and verified the token. ??? **We're also loading the resource but that is not required for :ensure_auth**.???

Hanspagh: In that sense its a user resource that is loaded by EnsureAuthenticated

Hanspagh:
@7stud any news on this?

Hanspagh:
Hi @7stud. We love to have this change, how is it coming along?

PR #609, opened 2019-09-02
tidies up token documentation

x-ji:
When using Guardian I found the documentation on tokens themselves to be relatively confusing. On many occasions I had to dig into Guardian source code to find the answers. I want to make it clearer.

- Expands JWT documentation
- Explains some default values used by `Guardian.Token.Jwt` in the moduledoc
- Fixes an error in moduledoc in `jwt.ex` which resulted in documentation not being rendered correctly
- Fixes some typos and errors

### DIFF
--- a/guides/tokens/start-tokens.md
+++ b/guides/tokens/start-tokens.md
@@ -4,16 +4,16 @@ Guardian uses the concept of a _token_ as a currency of authentication credentia
 
 Tokens can be anything that has the following properties:
 
-* tamper proof (signed/encrypted verifyable by the application)
+* tamper proof (signed/encrypted verifiable by the application)
 * include a payload (claims)
 
 Claims should be a map using string keys.
 
-JWT tokens (the default) fit these requirements and are widely supported in most languages using either hashed or cert based signing algorithms. However this is only one option. You can provide other behaviours like revocation, db tracking or another standard token type.
+[JWT tokens](https://en.wikipedia.org/wiki/JSON_Web_Token) (the default) fit these requirements and are widely supported in most languages using either hashed or cert based signing algorithms. However this is only one option. You can provide other behaviours like revocation, db tracking or another standard token type.
 
 Tokens can be provided by any means, they can be put into HTTP request headers, cookies, sessions; really as long as you can get your token to your application you can verify it and use it.
 
-Configuration of your token has two parts:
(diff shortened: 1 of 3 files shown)


guides/tokens/start-tokens.md
```diff
@@ -4,16 +4,16 @@ Guardian uses the concept of a _token_ as a currency of authentication credentia
...
 
 Claims should be a map using string keys.
 
-JWT tokens (the default) fit these requirements and are widely supported in most languages using either hashed or cert based signing algorithms. However this is only one option. You can provide other behaviours like revocation, db tracking or another standard token type.
+[JWT tokens](https://en.wikipedia.org/wiki/JSON_Web_Token) (the default) fit these requirements and are widely supported in most languages using either hashed or cert based signing algorithms. However this is only one option. You can provide other behaviours like revocation, db tracking or another standard token type.
 
 Tokens can be provided by any means, they can be put into HTTP request headers, cookies, sessions; really as long as you can get your token to your application you can verify it and use it.
 
-Configuration of your token has two parts:
+Configuration of your token comes from two places:
```
x-ji: I am not sure if I got the intention of this part correct, since the original sentence was left unfinished... If I got it wrong please let me know

lib/guardian/token/jwt.ex
```diff
@@ -119,7 +122,7 @@ defmodule Guardian.Token.Jwt do
   This will allow you to use the header values to determine dynamically the
   key that should be used.
 
-  ```
+  ```elixir
```
yordis: @x-ji thank you very much for the help.

Before we merge it, can we refactor the block of code markdown syntax from this page?

The block of codes should be 4 spaces of indentation without the triple backticks since this syntax is not actually supported by `ex_doc`.

x-ji: I think ExDoc actually supports it now. See https://github.com/elixir-lang/elixir/pull/5389

And there was a fenced code block in this same file just a few lines above, which seems to work fine.

Though of course I can also change it to be indented.

x-ji: In fact it seems to me that ```elixir is being used all over the project. You can just run a search over the project. So if I want to refactor this then we'll have to refactor the whole project. Unless I misunderstood something.

yordis:
🎉

PR #615, opened 2019-10-18
Tutorial Fixes

panbenson:


guides/tutorial/start-tutorial.md
```diff
@@ -102,7 +102,7 @@ defmodule AuthMe.UserManager.Guardian do
   end
 
   def resource_from_claims(%{"sub" => id}) do
-    case UserManager.get_user(id) do
+    case UserManager.get_user!(id) do
```
yordis: We use `!` when the function raises an error in Elixir. In this case, we are not using `rescue`.

I am curious why you suggested the change since it may be the case that we need to fix the tutorial somewhere else instead.

guides/tutorial/start-tutorial.md
```diff
@@ -444,5 +444,5 @@ Now exit iex and start up your server:
 $ mix phx.server
 ```
 
-Enter `localhost:4000/protected` in your browser's address bar, and you should see "unathenticated".  Now, enter `localhost:4000/login` in your browser's address bar, and login with your `me` username and `secret` password.  You should automatically be redirected to the protected page, which you can now see!  To logout, enter `localhost:4000/logout` in your browser's address bar, and you will be redirected to the login page.  Instead of logging in, enter `localhost:4000/protected` in your browser's address bar, and you will see "unathenticated" again!
+Enter `localhost:4000/protected` in your browser's address bar, and you should see "unauthenticated".  Now, enter `localhost:4000/login` in your browser's address bar, and login with your `me` username and `secret` password.  You should automatically be redirected to the protected page, which you can now see!  To logout, enter `localhost:4000/logout` in your browser's address bar, and you will be redirected to the login page.  Instead of logging in, enter `localhost:4000/protected` in your browser's address bar, and you will see "unauthenticated" again!
```
yordis: Is this line more than 80 characters width? Would you mind to break it down into multiple columns? It is really hard to figure what exactly changes.

guides/tutorial/start-tutorial.md
panbenson: sorry I'm quite new to elixir but changed to use `!` because the generator only created `get_user!`, otherwise function is undefined

guides/tutorial/start-tutorial.md
panbenson: there was just a typo in "unauthenticated" that I fixed! :)

guides/tutorial/start-tutorial.md
yordis: Ahhhh I see why you want it, in that case, changing it to `!` will no work thou.

Are you talking about Phoenix generator?

panbenson:
Yes the phoenix generator created that function. I managed to get it
running without errors using it however

Hanspagh:
Do you still want you other changes merged? @panbenson

panbenson:
@Hanspagh yeah sure, if you find the changes helpful, as those points were places i needed to change!

guides/tutorial/start-tutorial.md
Hanspagh: So it seems that phoenix generators removed the non-error returning version of get_uses, so we will have to use the get_user! method now, but we should also update the case statement to reflect this 

```suggestion
    case UserManager.get_user!(id) do
          {:error, err} -> {:error, :resource_not_found}
          {:ok, user} -> {:ok, user}
```

Hanspagh:
I have addeda bit more feedback

guides/tutorial/start-tutorial.md
```diff
@@ -102,7 +102,9 @@ defmodule AuthMe.UserManager.Guardian do
   end
 
   def resource_from_claims(%{"sub" => id}) do
-    case UserManager.get_user(id) do
+    case UserManager.get_user!(id) do
+          {:error, err} -> {:error, :resource_not_found}
+          {:ok, user} -> {:ok, user}
       nil -> {:error, :resource_not_found}
```
Hanspagh: You still have some problems here since the old clause has not been removed

panbenson:
cant be merged until @yordis approves requested changes 😭

PR #616, opened 2019-10-19
add SlidingCookie plug

mwri:
Adds a `SlidingCookie` plug which implements a sliding window/session behaviour, allowing the remember_me cookie to be replaced before it expires.

Configure (or use plug option) like so:

```elixir
  ttl: {10, :minutes},
  sliding_cookie: {5, :minutes}
```

Add to Guardian pipeline, for example:

```elixir
  plug Guardian.Plug.VerifyCookie
  plug Guardian.Plug.SlidingCookie
  plug Holco.Web.Plug.PostAuth
```

Provide `sliding_cookie` callback implementation:

```elixir
defmodule MyAppWeb.Guardian do
  def subject_for_token(user_rec, claims), do: ......
  def resource_from_claims(claims), do ......

  def sliding_cookie(claims, resource, opts) do
    {:ok, %{"new_claim" => "whatevs"}}
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,8 +297,8 @@ Look for a token in cookies and exchange it for an access token
 
 #### `Guardian.Plug.SlidingCookie`
 
-Replace a token in cookies with a new one if it is older than a configured
-threshold.
+Replace the token in cookies with a new one when a configured minimum TTL
+is remaining.
 
 #### `Guardian.Plug.EnsureAuthenticated`
 
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ conn = MyApp.Guardian.Plug.sign_in(conn, resource, %{some: "claim"}, ttl: {1, :m
 conn = MyApp.Guardian.Plug.sign_out(conn)
 
 # Set a "refresh" token directly on a cookie.
-# Can be used in conjunction with `Guardian.Plug.VerifyCookie`
+# Can be used in conjunction with `Guardian.Plug.VerifyCookie` and `Guardian.Plug.SlidingCookie`
 conn = MyApp.Guardian.Plug.remember_me(conn, resource)
 
 # Fetch the information from the current connection
@@ -294,6 +294,11 @@ Look for a token in the session and verify it
 #### `Guardian.Plug.VerifyCookie`
 
 Look for a token in cookies and exchange it for an access token
+
+#### `Guardian.Plug.SlidingCookie`
+
+Replace a token in cookies with a new one if it is older than a configured
+threshold.
 
 #### `Guardian.Plug.EnsureAuthenticated`
 
--- a/lib/guardian.ex
+++ b/lib/guardian.ex
@@ -473,6 +473,20 @@ defmodule Guardian do
       def exchange(token, from_type, to_type, opts \\ []),
         do: Guardian.exchange(__MODULE__, token, from_type, to_type, opts)
 
+      @doc """
+      If Guardian.Plug.SlidingCookie is used, this callback will be invoked to
+      return the new claims, or an error (which will mean the cookie will not
+      be refreshed).
+      """
+
+      @spec sliding_cookie(
+              current_claims :: Guardian.Token.claims(),
+              current_resource :: Guardian.Token.resource(),
+              options :: Guardian.options()
+            ) :: {:ok, new_claims :: Guardian.Token.claims()} | {:error, any}
+      def sliding_cookie(_current_claims, _current_resource, opts \\ []),
+        do: {:error, :not_implemented}
+
       def after_encode_and_sign(_r, _claims, token, _), do: {:ok, token}
       def after_sign_in(conn, _r, _t, _c, _o), do: {:ok, conn}
       def before_sign_out(conn, _location, _opts), do: {:ok, conn}
@@ -494,7 +508,8 @@ defmodule Guardian do
                      on_refresh: 3,
                      on_verify: 3,
                      peek: 1,
-                     verify_claims: 2
+                     verify_claims: 2,
+                     sliding_cookie: 3
     end
   end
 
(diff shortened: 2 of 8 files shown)


mwri:
Ah ha. Fixed formatting :)

mwri:
Rebased on master and added `:halt` plug option following merge of #617. I'm not sure it really ever makes sense not to halt when the implementation is broken, but I think it's nice to be entirely consistent with the other plugs.

lib/guardian.ex
```diff
@@ -767,6 +782,25 @@ defmodule Guardian do
     apply(mod, :config, [:token_module, @default_token_module])
   end
 
+  @doc false
```
Hanspagh: It might make sense to move this to https://github.com/ueberauth/guardian/blob/master/lib/guardian/token/jwt.ex and do a small refactor to avoid a bit of code dublication

README.md
```diff
@@ -295,6 +295,11 @@ Look for a token in the session and verify it
 
 Look for a token in cookies and exchange it for an access token
 
+#### `Guardian.Plug.SlidingCookie`
+
+Replace a token in cookies with a new one if it is older than a configured
```
Hanspagh: I am thinking this sounds a bit backward. Would it make sense that, what we configure is the minimum remaining time of the token, before we would refresh it? For me, this is mostly the way I think about refresh time 

I think the current way can cause some trouble if you change the ttl of the refresh token.
Say we have a refresh ttl of 7 days, and we want to refresh only when there is one day left, we would set the sliding ttl to 6 days. Changing the refresh ttl to 14 would cause the refresh to happen when there are still 8 days left.  

@mwri  @ueberauth/developers  Let me hear what you think

lib/guardian/plug/sliding_cookie.ex
```diff
@@ -0,0 +1,115 @@
...
+      case Keyword.get(opts, :sliding_cookie, module.config(:sliding_cookie)) do
+        nil ->
+          :no_sliding_window
+
+        ttl_descr ->
+          {:ok, ttl_to_seconds(ttl_descr)}
+      end
+    end
+
+    defp find_token_from_cookies(conn, opts) do
```
Hanspagh: Maybe we should share this function between verifyCookie and slidingCookie?

mwri:
Sorry, I don't know how I accidentally closed it there, must have hit a key when the button was highlighted or something.

lib/guardian.ex
mwri: Well.. it's funny you should say this, as a version of this logic did previously exist in jwt.ex, I necessarily decoupled the `Map.put` stuff in `assign_exp_from_ttl` from the actual ttl parsing to avoid the duplication, and then moved it out of jwt.ex to guardian.ex as I viewed it as a general utility function; it didn't seem to have anything really to do with the JWT token implementation at all as such at that point... and calling code from jwt.ex to parse a config option just didn't seem quite right to me... I'll move it to jwt.ex if you like though.

README.md
mwri: I'm absolutely with you on this, I tossed up about which way to go to be honest, a time after which it's renewed vs a time relative to the expiry... I'm very happy to change it to a minimum time remaining instead.

lib/guardian/plug/sliding_cookie.ex
mwri: Sure. I don't like code duplication but let this pass as it was a small function, and I wasn't sure it was worth coupling the two modules to share it. I don't feel strongly either way though, happy to call the `VerifyCookie` version.

mwri: Ah, could move it to plug.ex.

lib/guardian.ex
Hanspagh: I think you are right, that it should actually live in `guardian`, and we should just refactor `jwt` to use the code from `guardian` and not the other way around

README.md
Hanspagh: I think that would make sense :D

lib/guardian/plug/sliding_cookie.ex
Hanspagh: Would be great I think we have some code in `plug` that basically does this aswell

mwri: I looked at that, it has quite a lot in common with `fetch_token_key/2`, but it they don't really refactor together nicely IMO; I think the result would just be more abstract less clear code. The kicker is `find_token_from_cookies/2` does `token = conn.req_cookies[key] || conn.req_cookies[to_string(key)]`, using a string and atom version of the key. I'm not sure if this is necessary, the unit tests certainly pass fine with just token = conn.req_cookies[to_string(key)]` and my app is ok with it, but I'm a little reluctant to change such an obviously deliberate effort without more research, I don't know what behaviour older versions of Phoenix might have had that might necessitate it, or something else I haven't thought of...

mwri:
Agreed changes completed, I think.

Hanspagh:
Would love to have a view from one other @ueberauth/developers

lib/guardian/plug/sliding_cookie.ex
```diff
@@ -0,0 +1,107 @@
...
+      end
+    end
+
+    defp sliding_window(module, opts) do
+      case Keyword.get(opts, :sliding_cookie, module.config(:sliding_cookie)) do
+        nil ->
+          :no_sliding_window
+
+        ttl_descr ->
+          {:ok, ttl_to_seconds(ttl_descr)}
```
yordis: Wouldn't it be simpler to expect a number instead of a tuple?

If people want readability on what the unit of the number is they can always use `:timer` like `:timer.hours(2)` 🤷‍♂

mwri: I don't have much of an opinion as such on that, but I do think it's nice to be consistent and this is how the TTL config/option is specified.

mwri:
Rebased.

mwri:
Cheers. Always a pleasure to contribute a well received PR.

PR #617, opened 2019-10-19
feat: configure conn halt

hauptbenutzer:
Hi! 

I finally got around to putting this PR together. It addresses the discussion in #401. 
The `halt` option is added to all relevant plugs and defaults to true, as to not break any current behaviour. Let me know what you think.

Cheers!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v2.1.0 (unreleased)
+
+### Enhancement
+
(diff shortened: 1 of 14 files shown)


yordis:
@hauptbenutzer thank you very much. Would you mind adding a new entry to the CHANGELOG file following the previous examples?

hauptbenutzer:
@yordis Like this? Not really sure what an appropriate version bump would be.

yordis:
@hauptbenutzer yeah that was totally fine

PR #621, opened 2019-11-02
Version change to 2.0 in README

pikender:
I am little confused but looks like a typo.

In mix.exs, the version is `@version "2.0.0"` so should we not use `~> 2.0` in *Installation* section now.

Installation section has `~> 1.2`

```
# mix.exs

defp deps do
  [{:guardian, "~> 1.2"}]
end
```

Please ignore if I missed some hidden context here

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ mix.exs
 
 ```elixir
 defp deps do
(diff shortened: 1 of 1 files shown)


yordis:
Thank you!

PR #622, opened 2019-11-04
add auth_time and max_age features

mwri:
Introduces the `auth_time` config and equivalent option, which makes `auth_time` an essential claim, and the `max_age` config and equivalent option, which introduces enforcement of a maximum time since end user authentication. The `max_age` config is the same format as `ttl` and `sliding_cookie`. The `auth_time` is maintained cross token exchange, though of course creating a new token outside such a context will by default reset it to the current (iat) time.

Configuring `max_age: {24, :hours}`, which implies `auth_time: true` will mean `Guardian.Plug.SlidingCookie` will only be able to chain tokens for 24 hours before  verify is failed and end user authentication is mandated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,9 @@ Creating with custom claims and options
 # Customize the secret
 {:ok, token, claims} = MyApp.Guardian.encode_and_sign(resource, %{}, secret: "custom")
 {:ok, token, claims} = MyApp.Guardian.encode_and_sign(resource, %{}, secret: {SomeMod, :some_func, ["some", "args"]})
+
+# Require an "auth_time" claim to be added.
+{:ok, token, claims} = MyApp.Guardian.encode_and_sign(resource, %{}, auth_time: true)
 ```
 
 Decoding tokens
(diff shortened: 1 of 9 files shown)


Hanspagh:
I like the idea of this, and just wanted to ask if this specification is taken directly from the OpenID spec defined here https://openid.net/specs/openid-connect-core-1_0.html. If this is the case maybe it could make sense to somehow note that.

mwri:
Yes exactly, it's implemented per that spec. There is some orthogonal context, like **max_age** is configuration in this implementation, rather than a request in the spec, but I think a Guardian backed token service will naturally conform to the spec in an unsurprising way if **max_age** or **auth_time** is configured, or if the options are used according to some request parameters or other.

So, I think you're right, it's reasonable to reference the spec, but I'm not sure where particularly. I could add it to the README near the config documentation...

mwri:
Added link to OpenID Connect Core spec and rebased.

mwri:
You're very welcome.

Issue #607, opened 2019-08-27
Guardian.Permissions.Plug Issue when using one_of option


zacksiri:
```
** (Protocol.UndefinedError) protocol Enumerable not implemented for 0.
```

I'm getting this error when using the `Guardian.Permissions.Plug` inside my controller and using the `one_of` option the error is pointing to line 61 of the code which is 

https://github.com/ueberauth/guardian/blob/master/lib/guardian/permissions/plug.ex#L61

I downgraded to version 1.2 and everything is working normally.

Hanspagh:
Hi @zacksiri. Did you see that we had a breaking change when upgrading to version 2.0
https://github.com/ueberauth/guardian/blob/master/CHANGELOG.md#breaking-change

zacksiri:
@Hanspagh ah I see totally missed that thank you! I will try it out.

zacksiri:
@Hanspagh Hey I tried implementing the changes, I'm getting the same error. Actually before I saw the breaking changes I already read through the code to figure out to come to the same conclusion the only thing I didn't do was set the encoding option which I read in the code will set it to the BitwiseEncoding module by default already.

yordis:
@zacksiri would be possible to create a PR replicating the issue in the tests?

It will help us to debug based on your use case, and it will easier to do some testing around the issue as well.

Otherwise, post some code snippets that could help us to know how you are using the feature.

zacksiri:
@yordis understood, I will get to it.

Hanspagh:
@zacksiri. Did you have some code we could look at?

zacksiri:
@Hanspagh Hey no unfortunately I haven't had the time to create a sample project yet. I have downgraded to 1.2 for now since there are some projects I need to ship. I will upgrade my project again later to see if it still persists and will note down some code snippets.

Hanspagh:
@zacksiri Sure, I would really like to fix this thought, can you by any chance post what you have as params to `one_of`. Something like this asuming you use plug
```
plug Guardian.Permissions.BitwiseEncoding. one_of: [
    %{default: [:public_profile], user_actions: [:books]},
    %{default: [:public_profile], user_actions: [:music]},
  ]
```

nathany:
@Hanspagh Why is there a period after BitwiseEncoding in your example? Also in the documentation: https://github.com/ueberauth/guardian/blob/master/lib/guardian/permissions.ex#L88

```
[...]

PR #625, opened 2019-11-08
The default for missing bits is empty list

geofflane:
Zero as a default value was trying to be used in MapSet.new which is invalid
The normal return type is a list of atoms, so an empty list seems appropriate


geofflane:
This should fix #607

PR #626, opened 2019-11-30
Fix missing claims key of VerifySession and VerifyHeader plug tutorial.

tbson:


doomspork:
Welcome and thank you @tbson :tada:

PR #628, opened 2019-12-05
Change case statements to match updated phoenix generators

panbenson:
duplicate of stale PR https://github.com/ueberauth/guardian/pull/615
@Hanspagh could you please review this again so I can close this and my old PR?


guides/tutorial/start-tutorial.md
```diff
@@ -102,9 +102,9 @@ defmodule AuthMe.UserManager.Guardian do
   end
 
   def resource_from_claims(%{"sub" => id}) do
-    case UserManager.get_user(id) do
-      nil -> {:error, :resource_not_found}
-      user -> {:ok, user}
+    case UserManager.get_user!(id) do
```
doomspork: I believe we'll need to update this function a bit more as a result of shift to the bang function. 

In Elixir the convention is for non-bang (ex. `get_user/1`) to return `{:ok, result}` or `{:error, reason}`, for which this `case` would work perfectly. However, since Phoenix appears to no longer generate a non-bang we need to take into account the difference: these bang functions (`get_user!/1`) return `result` or raise an error; under-the-hood the Phoenix code uses [Repo.one!/1](https://hexdocs.pm/ecto/Ecto.Repo.html#c:one!/2).

To accommodate this difference I believe we'll want something like:

```elixir
def resource_from_claims(%{"sub" => id}) do
  user = UserManager.get_user!(id)
  {:ok, user}
rescue 
  e in Ecto.NoResultsError => {:error, :resource_not_found}
end
```

What are your thoughts @yordis, @Hanspagh?

doomspork:
@Hanspagh / @panbenson in the future if we have a PR blocked because of another team member's stale review we can dismiss it :+1: @Hanspagh you should have permissions to do so or you can ping me to take a peek.

Sorry about the last PR getting caught up in limbo @panbenson, we appreciate your interest in getting involved with the project :grin:

guides/tutorial/start-tutorial.md
Hanspagh: I agree with @doomspork, sorry for the misleading advise in the old PR

Hanspagh: I agree with @doomspork, sorry for the misleading advise in the old PR

PR #629, opened 2019-12-13
Updated start-tutorial

cgorshing:
I was walking through the tutorial and came across some issues. Pretty simple stuff. Let me know if you have any questions around these changes. 

Updated version of guardian

There isn't a `get_user` only a `get_user!` in the UserManager class

The changeset failed having `%User{}` ... `lib/auth_me/user_manager/user.ex:17: User.__struct__/0 is undefined, cannot expand struct Use`

The path for the code snippet was wrong (my first time through I created the file)

I updated the tests in the docs, this wasn't really an issue, but thought best to keep them consistent with what actually get generated.

### DIFF
--- a/guides/tutorial/start-tutorial.md
+++ b/guides/tutorial/start-tutorial.md
@@ -29,7 +29,7 @@ You'll need to update the dependencies to whatever is latest.
 
 defp deps do
   [
-    {:guardian, "~> 1.2"},
+    {:guardian, "~> 2.0"},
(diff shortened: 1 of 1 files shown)


doomspork:
Thank you @cgorshing!

@Hanspagh I am going to pull the commit from #628, amend the change, and push it so @panbenson's commits and contribution get it but we unblock this PR and are able to incorporate Benson's changes :+1:

PR #649, opened 2020-03-30
use exp/ttl in remember_me plug

### LINKED ISSUE #645: remember_me ignores token's TTL
Here we're trying to get `max_age` for a cookie from the token itself but because of the reason described in #570 (list concatenation) we ignoring the value and always using the value from config or defaults. Do we really need to have a default value for `max_age` (or allow it to be configured)? Why not simply rely on token's TTL itself? Or there is a reason for keeping a cookie alive longer than token's TTL?

https://github.com/ueberauth/guardian/blob/62744fb61fd62afe8bbb49b75b377be14aae65e6/lib/guardian/plug.ex#L318-L321

The existing behavior brakes the following use case:

```elixir
MyGuardian.Plug.remember_me(conn, resourse, claims, ttl: {90, :days})
```
Then one will expect a client get a refresh token for ~3 months but instead, the cookie will be expired after 28 days (which is current default `max_age`).

Of course, you can work around this and place into config `cookie_options: [max_age: 60 * 60 * 24 * 90]` but it feels wrong.

Hanspagh:
As described in #645 the current remember_me plug will use the default max_age or it can be overridden in the config. This pr will make the remember_me plug use the `exp` of the token if it's available as a default.
closes #645

### DIFF
--- a/lib/guardian/plug.ex
+++ b/lib/guardian/plug.ex
@@ -42,7 +42,7 @@ if Code.ensure_loaded?(Plug) do
     end
 
     @default_key "default"
-    @default_cookie_options [max_age: 60 * 60 * 24 * 7 * 4]
+    @default_cookie_max_age [max_age: 60 * 60 * 24 * 7 * 4]
 
     import Guardian, only: [returning_tuple: 1]
     import Guardian.Plug.Keys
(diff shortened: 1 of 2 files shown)


Hanspagh:
@ueberauth/developers bump

Hanspagh:
I would love some feedback on this @ueberauth/developers

Hanspagh:
🎉

PR #670, opened 2021-05-05
Allow refreshing tokens without typ field

urmastalimaa:
The standard "typ" field is not mandatory for JWTs. The token refresh
implementation does not actually rely on the "typ" field in case a
custom TTL is provided, but the function header mandates the field.
Instead of matching in the function header, move fetching the token type
in to the branch where it is actually necessary.


Hanspagh:
Hi @urmastalimaa 
Looking good, do you by any chance a reference to the docs for this?

urmastalimaa:
Hey @Hanspagh, can you clarify what docs are you referring to?

urmastalimaa:
Just in case, this is the "typ" field documentation in the JWT RFC: https://tools.ietf.org/html/rfc7519#section-5.1

Hanspagh:
Thank you. This was what I was looking for. This looks perfect.

Hanspagh (approved):

PR #675, opened 2021-06-28
chore: improve verify cookie options

### LINKED ISSUE #668: "token_expired" error handling uniformisation between VerifyCookie and VerifySession
Hello !

I’ve encountered an incoherence between `VerifyCookie` and `VerifySession` for a `token_expired` error.  
`VerifyCookie` ignores it whereas `VerifySession` does not and I think it complexifies some pipelines.
 This incoherence has been raised here https://github.com/ueberauth/guardian/issues/517

`VerifyCookie`:
https://github.com/ueberauth/guardian/blob/d4416658b613eae59aa833f3128bd16a8c557836/lib/guardian/plug/verify_cookie.ex#L85

`VerifySession`:
https://github.com/ueberauth/guardian/blob/d4416658b613eae59aa833f3128bd16a8c557836/lib/guardian/plug/verify_session.ex#L66

---

An exemple where this difference is problematic:
I implemented the [remember_me](https://hexdocs.pm/guardian/Guardian.Plug.html#remember_me/5) function in order to create a refresh token alongside a session cookie with my access token in it.
If my access token is expired, I would like to check if a refresh exists/valid to exchange it for an access token. 

To handle such case, I have to ignore the halt in `Guardian.Plug.VerifySession` which will allow `Guardian.Plug.VerifyCookie` to be called but also handle this error directly in my `error_handler` (could be handled in `ensure_authenticated` plug in some cases). 
```elixir
 plug(Guardian.Plug.VerifySession, claims: @claims, halt: false)
 plug(Guardian.Plug.VerifyCookie)
```

```elixir
def auth_error(conn, {:invalid_token, :token_expired}, _opts) do
  conn
end
```

  Wouldn’t be best to uniformise both plugs or at least provide an option ?
  Maybe my pipeline isn't coherent, feel free to tell me :) !

StephaneRob:
Try to fix #668.

No breaking changes, only deprecate `Guardian.Plug.VerifyCookie` in favor of a `:verify_cookie` on `Guardian.Plug.VerifySession` and `Guardian.Plug.VerifyHeader` plugs

### DIFF
--- a/test/guardian/plug/verify_header_test.exs
+++ b/test/guardian/plug/verify_header_test.exs
@@ -223,7 +223,9 @@ defmodule Guardian.Plug.VerifyHeaderTest do
         :get
         |> conn("/")
         |> put_req_header("authorization", ctx.token)
-        |> VerifyHeader.call(module: ctx.impl, refresh_from_cookie: [module: ctx.impl])
+        |> Pipeline.put_module(ctx.impl)
+        |> Pipeline.put_error_handler(ctx.handler)
+        |> VerifyHeader.call(refresh_from_cookie: [])
 
       assert Guardian.Plug.current_token(conn, []) == ctx.token
       assert Guardian.Plug.current_claims(conn, []) == ctx.claims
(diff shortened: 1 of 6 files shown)


Hanspagh:
This looks great. I will have a full look later today

test/guardian/plug/verify_header_test.exs
```diff
@@ -172,4 +172,96 @@ defmodule Guardian.Plug.VerifyHeaderTest do
...
+      {:ok, token, claims} = __MODULE__.ImplJwt.encode_and_sign(@resource)
+      {:ok, %{claims: claims, conn: conn(:get, "/"), token: token, impl: impl, handler: handler}}
+    end
+
+    test "when session is valid", ctx do
+      conn =
+        :get
+        |> conn("/")
+        |> put_req_header("authorization", ctx.token)
+        |> VerifyHeader.call(module: ctx.impl, verify_cookie: [module: ctx.impl])
```
Hanspagh, line 226: To avoid all the `module: ctx.impl` madness you can use `|> Pipeline.put_module(ctx.impl)` then we can also test the option true `to verify_cookie`, what do you think?

StephaneRob:
Hi @Hanspagh!
I updated Readme and module doc plugs. 
I agree, the name is not the best for this options. I've also changed it to `:refresh_from_cookie`.

Hanspagh:
Apart from the above, I don't have anything else @ueberauth/developers ?

test/guardian/plug/verify_header_test.exs
StephaneRob: I will add a test case for `refresh_from_cookie: true` (new `verify_cookie` :)).

StephaneRob:
@Hanspagh,  I added a test with `refresh_from_cookie: true` and clean tests by using ` Pipeline.put_module` and ` Pipeline.put_error_handler`

Hanspagh (approved):
Looks good. No more comments from me. 👍

PR #680, opened 2021-07-17
Add scheme option to VerifyHeader

### LINKED ISSUE #464: Incorrect / confusing use of the `:realm` option?
Hi,

First of all thanks for the great work on Guardian, it's a library that many of my colleagues and myself have been benefiting from immensely. :)

We have a use case where we need to support accepting one JWT from potentially different sources (issuers). After digging around and experimenting, we've figured out we can use the `:realm` option to differentiate the JWTs so that the token verification works as intended.

However, we noticed that the `:realm` option in `Guardian.Plug.VerifyHeader` does not conform to any of the RFCs<sup>1</sup>:

The default value for `:realm` is `Bearer`, but according to the RFCs and [IANA](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml), `Bearer` is a type of HTTP authentication scheme, whereas `realm` (and others) are attributes.

To accept a Bearer token with a particular realm, in Guardian we have to work around it by setting it like: `realm: "Bearer realm=\"example.com\""`.

Ideally we should have a `:scheme` option that defaults to `Bearer`, and either a dedicated `:realm` option, or a more flexible `:attributes` option that allows any attributes.

I am in no way an expert on this topic, so please let me know if I've missed something. Thanks!

---

1. RFCs I've looked into:

- RFC2617 HTTP Authentication: https://tools.ietf.org/html/rfc2617
- RFC6749 OAuth 2.0 https://tools.ietf.org/html/rfc6749
- RFC6750 OAuth 2.0 Bearer Token https://tools.ietf.org/html/rfc6750
- RFC7235 HTTP/1.1 https://tools.ietf.org/html/rfc7235
- RFC7519 JWT https://tools.ietf.org/html/rfc7519

yordis:
closes #464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## v2.2.0
+
+### Enhancement
+
+* Add `:scheme` option to `Guardian.Plug.VerifyHeader` [#680](https://github.com/ueberauth/guardian/pull/680)
+
+### Deprecation
+
+* `:scheme` option configuration of  `Guardian.Plug.VerifyHeader` is deprecated
+  please use `:scheme` instead.
+
 ## v2.1.2
 ### Enhancement
 * Documentation improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Deprecation
 
-* `:scheme` option configuration of  `Guardian.Plug.VerifyHeader` is deprecated
+* `:realm` option configuration of  `Guardian.Plug.VerifyHeader` is deprecated
   please use `:scheme` instead.
 
 ## v2.1.2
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
 
+## v2.2.0
+
+### Enhancement
+
+* Add `:scheme` option to `Guardian.Plug.VerifyHeader` [#680](https://github.com/ueberauth/guardian/pull/680)
+
+### Deprecation
+
+* `:scheme` option configuration of  `Guardian.Plug.VerifyHeader` is deprecated
```
doomspork, line 11: This must be a typo, scheme is being replaced by scheme?

doomspork (changes requested):

doomspork (approved):

PR #682, opened 2021-08-11
Allow multiple `resource_from_claims/1` signatures in the README example

thepeoplesbourgeois:
The way the `subject_for_token/1` and  `resource_from_claims/1` examples in the readme are currently written, no call would ever reach their error paths, as there is no pattern to match against in each preceding signature, and those would therefore act as the catchall case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,13 +57,13 @@ To do this, create a module that uses `Guardian` and implements the `subject_for
 defmodule MyApp.Guardian do
   use Guardian, otp_app: :my_app
 
-  def subject_for_token(resource, _claims) do
+  def subject_for_token(%{id: id}, _claims) do
     # You can use any value for the subject of your token but
     # it should be useful in retrieving the resource later, see
     # how it being used on `resource_from_claims/1` function.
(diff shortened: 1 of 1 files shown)


thepeoplesbourgeois:
Temporarily closing, will reopen shortly

yordis (approved):

yordis:
Thank you so much, would you mind adding an entry here: https://github.com/ueberauth/guardian/blob/master/CHANGELOG.md

thepeoplesbourgeois:
It would be the first such entry of its kind, according to a quick "find on page" for the word "readme"... perhaps save the changelog for updates to how the package and its API work? 🙂

thepeoplesbourgeois:
My apologies; I was considering the PR language in terms of how its changes would appear within the blame records of the affected file. I should have indicated somewhere in the name or details of the PR that this is a change to examples in the README file. I've done so now

yordis:
It is OK, I should figure out how to make it better for you'all to help us, don't worry about it.

PR #683, opened 2021-08-17
fix: try refreshing cookie on multiple cases

StephaneRob:
Currently we try to refresh cookie only when the access token is found and is expired.
If the refresh cookie is present we should try refreshing session with it even if invalid or not found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v2.2.1
+
+### Enhancement
+
+* `Guardian.Plug.VerifyHeader` and `Guardian.Plug.VerifySession` `:refresh_from_cookie` option will try refreshing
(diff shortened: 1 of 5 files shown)


yordis:
Would you mind adding an entry in the changelog https://github.com/ueberauth/guardian/blob/master/CHANGELOG.md

yordis (changes requested):
Please add some entry in the changelog so we keep the PR ready to be release

StephaneRob:
Hi @yordis, I added some test cases and updated the CHANGELOG.
The changelog doesn't mention the neither `refresh_from_cookie` option nor `VerifyCookie` plug deprecation. Do you want I add it under `2.2.0` version?

yordis:
@StephaneRob please if you don't mind I would appreciate updating the CHANGELOG for sure

StephaneRob:
@yordis done.

yordis:
You are the best! Thank you so much!

yordis (approved):

PR #688, opened 2021-12-23
Corrects Plug.Verify presentation on pipelines.md

kimjoaoun:
When presenting the function `Guardian.Plug.VerifyCookie`, the documentation says: "A cookie has the cookie stored". It is supposed to say "A cookie has the token stored".


yordis (approved):

yordis:
💜

PR #692, opened 2022-02-25
Minor docs improvements, mostly around Verify* plugs

alopezz:
What prompted me to work on this was that as I was trying to add some custom verification logic, it wasn't obvious to me from the documentation that the `Verify*` plugs would let a session without a token (e.g. a logged out session) pass, and that `EnsureAuthenticated` is typically used for little more than checking that a token is present.

So this tries to be more clear on that, and takes the opportunity to do some other minor fixups as well.

### DIFF
--- a/guides/plug/pipelines.md
+++ b/guides/plug/pipelines.md
@@ -2,9 +2,9 @@
 
 Guardian's Plug support provides an easy and composable way to put together your authentication.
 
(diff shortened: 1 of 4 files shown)


yordis:
Hey there, my apologies for the delay, thank you so much for your help.

Do you mind rebasing?

alopezz:
Rebased just now.

yordis (approved):
Thank you so much 🚀

PR #696, opened 2022-03-17
Permit atom keys when verifying claims with EnsureAuthenticated

giddie:


yordis:
Hey peep, do you mind sharing what is the use case for the feature? Not sure what the limitations were before.

giddie:
Sure, it's pretty simple really:

```elixir
# This halts to conn, even when claims match `%{"custom_claim" => true}`
plug Guardian.Plug.EnsureAuthenticated, claims: %{custom_claim: true}

# This works as expected
plug Guardian.Plug.EnsureAuthenticated, claims: %{"custom_claim" => true}
```

Since other functions, such as `decode_and_verify` handle atom keys OK when verifying claims, I was a bit confused by this behaviour.

yordis:
Do you mind adding an entry in the `CHANGELOG.md` and bumping the version accordingly?

yordis (approved):

PR #697, opened 2022-03-24
Ensure that badly-formatted tokens don't raise an exception

giddie:
It seems that `JOSE` likes to raise `ArgumentError`s when it encounters badly-formatted tokens. Since Guardian returns an `:ok | :error` tuple when decoding tokens, it seems sensible to catch these exceptions where possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v2.2.3
+
+### Enhancement
+
+* Ensure that badly-formatted tokens don't raise an exception when attempting to decode them.
+
 ## v2.2.2
 
 ### Enhancement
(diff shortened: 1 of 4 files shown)


yordis:
Hey peep, do you mind bumping the version of the package and adding a CHANGELOG entry as well?

yordis (approved):

yordis:
Any clue why the CI is not running?

giddie:
Nope, no idea :shrug:

dolfinus:
Hi.

It's better to change version from 2.2.2 to 2.3.0, because this could cause backward compatibility issues.

yordis:
@dolfinus blame me for that one, I will do my best the next time.

Issue #701, opened 2022-06-28
use Bitwise is deprecated in elixir 1.14-dev


shahryarjb:
### Steps to Reproduce

Hi friends, I test my code in elixir 1.14-dev before releasing official condition

### Expected Result

..

### Actual Result

```elixir
warning: use Bitwise is deprecated. import Bitwise instead
  lib/guardian/permissions/atom_encoding.ex:7: Guardian.Permissions.AtomEncoding

warning: use Bitwise is deprecated. import Bitwise instead
  lib/guardian/permissions/bitwise_encoding.ex:7: Guardian.Permissions.BitwiseEncoding

warning: use Bitwise is deprecated. import Bitwise instead
  lib/guardian/permissions/text_encoding.ex:7: Guardian.Permissions.TextEncoding
```

PR #702, opened 2022-08-10
Fix deprecation warnings based on 1.14-dev.

crova:
Greetings, how is it going?

As [mentioned on this issue](https://github.com/ueberauth/guardian/issues/701) there are a few `warnings` when compiling while using elixir `1.14`.

Besides those that were mentioned, when `use`ing `Guardian` on a module, the following warn would raise:
```elixir
warning: Application.get_env/3 is discouraged in the module body, use Application.compile_env/3 instead
```

This PR address both issues.

Please let me know if I should improve this PR somehow. 

Cheers.


lib/guardian.ex
```diff
@@ -343,7 +343,7 @@ defmodule Guardian do
       # Provide a way to get at the configuration during compile time
       # for other macros that may want to use them
       @config fn ->
-        the_otp_app |> Application.get_env(__MODULE__, []) |> Keyword.merge(the_opts)
+        the_otp_app |> Application.compile_env(__MODULE__, []) |> Keyword.merge(the_opts)
```
yordis, line 346: This is weird thou because this is not the module scope. The function is inside an anonymous closure.

Is this a false promise?

crova: Hey @yordis , I'm not sure I understand what you mean by a `false promise` here?!

We started getting this warning from a module within our application that just has `use Guardian, otp_app: :our_app`.
This was the line getting failing the compilation when running `mix compile --warnings-as-errors`.

The rest of the module is just some wrapper functions around `Guardian.DB.on_refresh` and such.

Let me know if I can provide more info here.

Cheers.

gungorkocak:
I'm also trying to upgrade Elixir v1.14.0-rc.1 and facing the same deprecation warning with `Application.get_env/3`.

Also `mix compile --force --warnings-as-errors` is a requirement for me(i.e. CI pipeline).
That ties upgrading Elixir to this PR (or a better solution).

lib/guardian.ex
yordis: I checked with Jose about it, and we are safe to go.

Let me do something else before we merge this one first 🙏🏻

crova: Let me know if you need help with anything else here @yordis .

Cheers.

kevinkirkup:
Related to #704

lib/guardian.ex
yordis: I recently added the following https://github.com/ueberauth/.github/blob/master/SECURITY.md

I am upgrading all the packages under Ueberauth to follow the supported versions. A lot of grunt work 😭 

We have a misleading setup at the moment and want to ensure people don't rely on unmaintained Elixir versions and whatnot.

Currently, I have an aggressive deadline, so I am trying my best to little by little do them.

What to do?

1. Update the `mix.exs` to reflect the supported Elixir versions
2. Bump the `mix.exs` to reflect such changes
3. Update the CHANGELOG.md entry
4. Create PR
5. Release the package
6. Repeat for every package

yordis: @crova do you mind following the https://github.com/ueberauth/.github/blob/master/SECURITY.md and the steps I shared before?

yordis (approved):

yordis:
💜

PR #712, opened 2022-10-18
Fix deprecation warnings based on 1.14

akiehm-ggs:
Similar to this PR: #702, when adding `use Guardian, otp_app: :api` to a module, we see the same warning:

`warning: Application.get_env/3 is discouraged in the module body, use Application.compile_env/3 instead`

Also `mix compile --force --warnings-as-errors` is a requirement for our CI pipeline.

### DIFF
--- a/lib/guardian/plug/pipeline.ex
+++ b/lib/guardian/plug/pipeline.ex
@@ -145,7 +145,7 @@ if Code.ensure_loaded?(Plug) do
 
             otp_app ->
               otp_app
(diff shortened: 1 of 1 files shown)


yordis:
I am so confused. That doesn't sound like the same code to me thou. Are you sure this will allow you to do the same as before?

akiehm-ggs:
After further investigation, we are no longer seeing the issue. Thanks for the quick reply!

PR #714, opened 2022-11-09
doc: Clarify session storage behavior on sign_in in README

KaylaBrady:
This clarifies that only the token is stored on the session when `Guardian.Plug.sign_in` is called - the claims and resource are not. See implementation [here](https://github.com/ueberauth/guardian/blob/d46795513fc36938e563fa2fc601392e4871baaf/lib/guardian/plug.ex#L209).

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,8 +118,8 @@ Congrats! We have a working Guardian implementation.
 With Plug:
 
 ```elixir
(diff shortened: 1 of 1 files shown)


yordis (approved):

yordis:
💜 thank you!

Issue #706, opened 2022-09-06
Unable to set secret_key in runtime.exs


geofflane:
I've assumed that setting secret_key was something that you would want to be secret and so am using the "12 Factor" style for pulling in that value to the config at runtime instead of committing a hard-coded value to the config.exs in the repo. The recent change #702 changed the reading of all of those config values to compile time which breaks the setting of that value at runtime.

It feels like all of those things, including TTL and Issuer you might want to vary by environment for various reasons and should be read at runtime and not at compile time.  

### Steps to Reproduce

1. Have a default value for your secret_key in config/config.exs
2. Override that value in config/runtime.exs

### Expected Result

Runtime value is used for secret, TTL, etc as needed.

### Actual Result

ERROR! the application :photon has a different value set for key PhotonPerm.Guardian during runtime compared to compile time. Since this application environment entry was marked as compile time, this difference can lead to different behaviour than expected:

* Compile time value was set to: [issuer: "PFF.Photon", ttl: {14, :days}, secret_key: "not secret, override in prod.exs"]
* Runtime value was set to: [issuer: "PFF.Photon", ttl: {14, :days}, secret_key: "xxxxxxx"]

To fix this error, you might:
* Make the runtime value match the compile time one
* Recompile your project. If the misconfigured application is a dependency, you may need to run "mix deps.compile photon --force"

Hanspagh:
@geofflane. As I read the code, if you omit the secret_key from your config at compile time, Guardian should correctly fetch it from the env at runtime. The compile time envs are a utility for the macros to use at compile time. With the change to use compile_env elixir now `tools like Mix will store the values used during compilation and compare the compilation values with the runtime values whenever your system starts, raising an error in case they differ.` Would mind trying to not set the secret key at compile time and only a runtime.

geofflane:
```
# config/config.exs
config :photon, PhotonPerm.Guardian,
  issuer: "PFF.Photon"
```
with
```
# config/runtime.exs
config :photon, PhotonPerm.Guardian,
  ttl: {14, :days},
  secret_key: System.fetch_env!("SECRET_KEY_BASE")
```

Has the same issue:
```
* Compile time value was set to: [issuer: "PFF.Photon"]
* Runtime value was set to: [issuer: "PFF.Photon", ttl: {14, :days}, secret_key: "xxx"]
```

idyll:
This is impacting us as well and I don't think the changes that were made are actually intended...

Just spelunking in the code a bit here:

[...]

PR #715, opened 2022-11-23
Users should be able to change most config values at runtime

geofflane:
Permissions are used at compile time, but other config variables should be read at runtime so that they can be overridden at runtime vs compile time. This allows for 12 Factor type deployment

This is intended to fix: #706 

Existing tests pass, but I'm not really sure how to write a test for this exact issue.
* `Application.put_env` in a test doesn't trigger the error. 
* Tried adding a test.exs and a runtime.exs with different values but couldn't trigger it either.
 
Happy to get some help on how to test that and update this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v2.3.1
+
+* Change compile time loading of configuration to only load permissions
+  allowing the app to change things like ttl or secret key at runtime
+
 ## v2.3.0
 
 * Fix warning about the usage of `Application.get_env` in the module scope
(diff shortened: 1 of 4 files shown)


yordis:
Hey, do you mind adding an entry here https://github.com/ueberauth/guardian/blob/master/CHANGELOG.md and changing the `mix.exs` version?

yordis (changes requested):
Waiting for the changelog updates.

Thank you so much for the work.

yordis:
@geofflane did you get to test this in your app at least? I agree it is a bit tricky to test for sure.

geofflane:
> did you get to test this in your app at least? I agree it is a bit tricky to test for sure.

Yes, I did change locally and 2.3.0 fails to start, but with this change the app will start with a changed secret_key value.

yordis (approved):

yordis:
💜 🚀

PR #727, opened 2024-04-01
header_name should always be downcase

gabeio:
get_req_header will never match for uppercase headers.

If a header_name is provided and is uppercase it will never match the header. Instead of simply documenting it, we should just assert that the header we are looking for is always downcase to prevent footguns.

### DIFF
--- a/lib/guardian/plug/verify_header.ex
+++ b/lib/guardian/plug/verify_header.ex
@@ -152,7 +152,9 @@ if Code.ensure_loaded?(Plug) do
             :no_token_found
             | {:ok, String.t()}
     defp fetch_token_from_header(conn, opts) do
-      header_name = Keyword.get(opts, :header_name, "authorization")
+      header_name =
+        Keyword.get(opts, :header_name, "authorization")
+        |> String.downcase
       headers = get_req_header(conn, header_name)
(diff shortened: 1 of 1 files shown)


yordis:
Hey there, thank you very much for your contribution!

Have you try to test what would happen if you pass a header with uppercase letter without your changes?

yordis:
Hey @gabeio,

Hope you're doing great! It's an interesting point because, within our ecosystem, `Plug` hasn't set a precedent for handling header casing. In my view, maybe `Guardian` should follow suit and not take on this responsibility either. This part of the setup is more or less set, and forget it, right? So, with thorough testing, we should be covered.

I see where you're coming from and acknowledge it's a potential hiccup. However, it hasn't popped up as a big issue so far. I'm hesitant to dive into defensive programming for something that isn’t a widespread problem, especially if `Plug` isn't going down that road either.

I'm not saying you're off base—I'm just offering another perspective. It's always great to discuss these things together. What do you think?

gabeio:
That's fair then, I'm new to the elixir scene. Then it might be best if we change this to documentation instead and just hint that the system which handles headers _might_ automatically lowercase the headers and that the user will need to determine that on their own.

PR #732, opened 2024-06-24
chore: use erlef actions

yordis:


.github/workflows/ci.yml
```diff
@@ -6,7 +6,53 @@ on:
   push:
     branches:
       - master
+
 jobs:
+  job1:
```
doomspork, line 11: Is this still a WIP?

yordis: yeah yeah, do not worry about this one, this is purely using as an experiment

PR #735, opened 2024-10-09
refactor: deprecate use of unless

johnmcguin:
As I was implementing Guardian in my project, I noticed the use of `unless/2` in the codebase. I had heard that the language was deprecating `unless/2` moving forward in 1.18 and later as verified [here](https://github.com/elixir-lang/elixir/pull/13851)


yordis (approved):

yordis:
🚀 💜

PR #738, opened 2025-07-02
Updates repo

michelboaventura:
I've updated all dependencies, replaced the old credo config to the newest version, fixed some lints and changed CI code to run against multiple erlang/elixir versions. For Erlang we are testing against 25, 26, 27 and 28. For Elixir 1.17 and 1.18


.github/workflows/ci.yml
```diff
@@ -9,39 +9,55 @@ on:
...
-        uses: actions/checkout@v1
+        uses: actions/checkout@v4
 
       - name: Set up Elixir
         uses: erlef/setup-beam@v1
         with:
-          version-file: .tool-versions
-          version-type: strict
+          otp-version: ${{matrix.otp}}
+          elixir-version: ${{matrix.elixir}}
```
doomspork, line 60: Does linting really need to be run across all of the combinations? I don't believe so 🤔

doomspork:
Thank you @michelboaventura! 

@ueberauth/developers we currently have Test as a required gate for PRs. With the matrix addition we probably need to decide what combinations are required and weren't are not. Thoughts?

yordis:
We should follow the exact same setup as the BEAM Community common-config as much as we can

doomspork (approved):

PR #745, opened 2026-08-16
Reject non-numeric JWT time claims

patric-vinicios:
I found this while testing Guardian tokens used for both HTTP requests and aPhoenix socket

A signed token with `exp` set to `"tomorrow"` was accepted by `decode_and_verify/1`. I checked `nbf` and `auth_time` as well and found the same behavior

The fallback in `time_within_drift?/2` returns `true` when the value is not a number. Because of that, malformed time claims can pass verification without going through the normal timestamp checks

This adds type checks for the time claims Guardian already verifies. Integers and floats keep working as before, while other values return `:invalid_token`. The existing behavior for `nil` is unchanged

I added tests for `exp`, `nbf`, `auth_time`, and the drift helper. The full test suite passes with 245 tests

### DIFF
--- a/lib/guardian/token/jwt/verify.ex
+++ b/lib/guardian/token/jwt/verify.ex
@@ -28,6 +28,9 @@ defmodule Guardian.Token.Jwt.Verify do
 
   def verify_claim(_mod, "nbf", %{"nbf" => nil} = claims, _opts), do: {:ok, claims}
 
+  def verify_claim(_mod, "nbf", %{"nbf" => nbf}, _opts) when not is_number(nbf),
+    do: {:error, :invalid_token}
+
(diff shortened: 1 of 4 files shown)


yordis (approved):

yordis:
CI failing

patric-vinicios:
> CI failing

I checked again and the second attempt failed in `Guardian.Permissions.AtomEncodingTest` which is outside this PR. I can reproduce the same error on the commit (`8ca4373`) without these changes.
The last attempt failed during `setup-beam` because it couldn't fetch the OTP build list from `builds.hex.pm` before dependencies were installed or teste were started

Both modified test files pass locally with 57 tests and also formatter and Credo pass

